PR #48, opened 2022-07-12
[WIP] reworked distances

AlekseySh:


DaloroAT:
Let's also add `pairwise_dist` and remove all calling of `cdist` in repo?

oml/utils/misc_torch.py
```diff
@@ -10,12 +10,26 @@
...
+    """
+    Args:
+        x1: tensor with the shape of [N, D]
+        x2: tensor with the shape of [N, D]
+        p: degree
+
+    Returns: elementwise distances with the shape of [N]
+
+    """
+    assert len(x1.shape) == len(x2.shape) == 2
```
DaloroAT, line 23: add extra assert for shape

AlekseySh: done

AlekseySh:
I think using cdist for pairwise is okay since in explicitly said in its description 

`    r"""Computes batched the p-norm distance between each pair of the two collections of row vectors. `

DaloroAT (approved):

DaloroAT (approved):

PR #50, opened 2022-07-13
Metrics refactoring

AlekseySh:


tests/test_examples/configs/train_mock.yaml
```diff
@@ -55,6 +55,10 @@ sampler:
 
 bs_val: 2
 
+metric_args:
+  cmc_top_k: [1, 5]
```
DaloroAT, line 59: add precision to val and train mock

oml/lightning/entrypoints/validate.py
```diff
@@ -11,9 +11,14 @@
 from oml.lightning.modules.retrieval import RetrievalModule
 from oml.metrics.embeddings import EmbeddingMetrics
 from oml.registry.models import get_extractor_by_cfg
+from oml.utils.misc import dictconfig_to_dict
 
 
 def main(cfg: TCfg) -> Tuple[pl.Trainer, Dict[str, Any]]:
+    print(cfg)
```
DaloroAT, line 18: hydra config should be resolved before print, so use `print` after `dictconfig_to_dict(cfg)`

DaloroAT (dismissed):

PR #56, opened 2022-07-13
Experiments

AlekseySh:


b0nce (approved):

AlekseySh:
added adamw; added best config for deepfashion; minor refactoring

PR #67, opened 2022-07-27
[WIP] Hard miner improvement

DaloroAT:
Added top miners, which form triplets from several hard positive and hard negative


AlekseySh:
moved to https://github.com/OML-Team/open-metric-learning/pull/202

PR #99, opened 2022-08-16
added scheduler args to config

AlekseySh:


oml/lightning/modules/retrieval.py
```diff
@@ -17,9 +17,9 @@ def __init__(
         scheduler: Optional[_LRScheduler] = None,
         scheduler_interval: str = "step",
```
b0nce, line 18: Remove this.

AlekseySh: discussed offline

b0nce (approved):

PR #100, opened 2022-08-16
Add ReduceLROnPlateau scheduler

AlekseySh:


examples/inshop/configs/train_inshop.yaml
```diff
@@ -2,9 +2,7 @@ postfix: metric_learning
 
 seed: 42
 precision: 32
-gpus:
-  - 0
-  - 1
+gpus: [0, 1]
```
AlekseySh, line 5: since you change it in only one config, please, rm it

by the way, @DaloroAT will change handling this arguments in the next PR, so you don't need to care about it

oml/configs/scheduler/reduce_on_plateau.yaml
```diff
@@ -0,0 +1,4 @@
+name: reduce_on_plateau
+args:
+  factor: 0.75
+  patience: 1
```
AlekseySh, line 4: this red circle indicates no new line in the end of the file, add it please

b0nce (approved):
Wow, my PR seems LGTM.

PR #112, opened 2022-08-22
[WIP] Fix DataLoaders + Samplers so they can work with DDP correctly

DaloroAT:


DaloroAT:
Previously we used DDP in training loop, but launched Trainer with flag `replace_sampler_ddp=False`. It's figure out that Lightning uses loader as is, without splitting them into several parts according the number of devices.

In this PR I tried to implement converting loaders to DDP version. In DDP mode devices should work with the same code, but with different part of data. 

Changelog:
0) It is necessary to ensure the launch of the same code depending on the presence or absence of DDP. 
1) MetricValCallback expect the number of samples for accumulation. In DDP this callback should expect less amount of data and after collecting data from different device, it is necessary to merge predicts from all devices, and then compute metrics. Values of metrics for experiments with different amount of devcies shouldn't be the equal, but they have to be similar, especially with large amount of data.
2) Process group for DDP with PL is creating only after launch `trainer.fit()` or `trainer.validate()`. So we need to replace loaders in runtime. For this purposes we added `ModuleDDP` which handle this replacement.
3) We want to replace `sampler` (validation, or not triplets pipeline for training in future) and `batch_sampler` (training) to DDP version. We added patcher for both of cases.
4) We want to be sure, that user will utilize DDP in correct way. So we added a lot of errors for wrong use cases.
5) We added a possibility to run code in DDP with cpu and gpu. We use it for testing.
6) Sequential samplers was removed and custom samplers have behaviour of batch sampler.

PR #136, opened 2022-09-08
Checkpointing parametrization with metric name.

b0nce:


AlekseySh (commented):
1 - raise instead of update
2 - check log_only_main_category
3 - rm regexp
4 - add few tests

oml/utils/misc.py
```diff
@@ -113,14 +112,28 @@ def clip_max(arr: Tuple[int, ...], max_el: int) -> Tuple[int, ...]:
     return tuple(min(x, max_el) for x in arr)
 
 
+def get_monitor_and_update_cfg(cfg: TCfg) -> str:
+    metric_args = cfg.setdefault("metric_args", {})
+    metric_for_ckpt = metric_args.get("metric_for_checkpointing", f"{OVERALL_CATEGORIES_KEY}/cmc/1")
+
+    metric_name, metric_topk = re.findall(r"([a-zA-Z]+)/(\d+)", metric_for_ckpt)[-1]
```
AlekseySh, line 119: @DaloroAT I guess we should remove it and use str.split(/), what do you think?

AlekseySh (approved):
lgfm

PR #148, opened 2022-09-19
Dropping last batch for BalanceSampler and reworked tests for samplers 

AlekseySh:


AlekseySh:
Changelog:
- reworked tests for samplers: old tests always produced epochs with the size of 1 batch
- removed `batches_in_epochs` from samplers and renamed few fields

b0nce:
Maybe we should force sampler to have `n_labels`, `n_instances` through interface if we are accounting for them in tests?

AlekseySh:
@b0nce don't think so because during the test we already know which kind of sampler we are testing since they behave in a different ways (so we have no `generic` tests when such an interface may be useful)

tests/test_oml/test_samplers/test_balance.py
```diff
@@ -66,64 +66,46 @@ def input_for_balance_batch_sampler() -> TLabelsPK:
...
-        num_labels_in_data * n_instances,
-    )
+
+    assert len(sampler) == n_expected_batches, (len(sampler), n_expected_batches)
+    assert (num_labels_in_data - sampler.n_labels + 1) <= num_labels_sampled <= num_labels_in_data
+    assert len(sampler) * sampler.n_labels * sampler.n_instances == n_instances_sampled
 
 
-def test_balance_batch_sampler(input_for_balance_batch_sampler: TLabelsPK) -> None:
+def test_balance_batch_sampler(input_for_balance_batch_sampler: TLabels) -> None:
```
DaloroAT, line 101: [suggestion] I prefer to launch tests with some parametrize, e.g. generate cases with different num of labels. Sometimes it generate you some cases with dividing, rounding, etc. 

Up to you

tests/test_oml/test_samplers/test_category_balance.py
```diff
@@ -202,39 +205,34 @@ def test_category_balance_batch_sampler(
...
+
         sampler = CategoryBalanceSampler(
             labels=labels,
             label2category=label2category,  # type: ignore
-            resample_labels=True,
-            n_categories=n_categories,
             n_labels=n_labels,
             n_instances=n_instances,
+            n_categories=n_categories,
+            resample_labels=True,
```
DaloroAT, line 230: parametrize True/False ?

DaloroAT (approved):

tests/test_oml/test_samplers/test_category_balance.py
AlekseySh: discussed offline

PR #155, opened 2022-09-27
Fixed building for pip 

AlekseySh:


AlekseySh:
seems like the problem was in finding packages: `packages=find_packages(include="oml")`

DaloroAT (approved):

PR #158, opened 2022-09-30
Add new storage for checkpoints and mock_dataset

DaloroAT:


AlekseySh:
Change log 
- now we download 4x vit weights from our storage instead of google drive
- we download the mock dataset from both our storage and google drive 
- removed short imports due to the problems with circular imports

AlekseySh (approved):

PR #233, opened 2022-11-22
FNMR@FMR metric.

dapladoc:
Add `calc_fnmr@fmr` function to functional/metrics.py and its call to `calc_retrieval_metrics`.
Add new parameter `fmr_vals` to `calc_retrieval_metrics`, `EmbeddingMetrics` and cofig files.
FNMR@FMR could't be used for data retrieval visualiztion, so it is added to `metrics_to_exclude_from_visualization` by default.
Add test for `calc_fnmr@fmr` and modify some other tests to work correctly with FNMR@FMR.

I had three test failed: `test_readme_was_built_correctly`, `test_code_blocks_in_readme` and `test_reduce_metrics` which is marked as it could be failed. The problem with `test_code_blocks_in_readme` is that I have just one GPU, so maybe it worth to be marked as it could be failed too. And as for `test_readme_was_built_correctly` I don't know what's wrong, because I didn't change readme files.

### DIFF
--- a/examples/cars/configs/train_cars.yaml
+++ b/examples/cars/configs/train_cars.yaml
@@ -37,6 +37,7 @@ metric_args:
   metrics_to_exclude_from_visualization: [cmc,]
   cmc_top_k: [1]
   map_top_k: [5]
+  fmr_vals: [0.5, 0.1, 0.01, 0.001]
   return_only_main_category: True
   visualize_only_main_category: True
 
--- a/examples/cars/configs/train_cars.yaml
+++ b/examples/cars/configs/train_cars.yaml
@@ -37,7 +37,7 @@ metric_args:
   metrics_to_exclude_from_visualization: [cmc,]
   cmc_top_k: [1]
   map_top_k: [5]
-  fmr_vals: [0.5, 0.1, 0.01, 0.001]
+  fmr_vals: [1]
   return_only_main_category: True
   visualize_only_main_category: True
 
--- a/oml/functional/metrics.py
+++ b/oml/functional/metrics.py
@@ -9,7 +9,7 @@ from oml.losses.triplet import get_tri_ids_in_plain
 from oml.utils.misc import clip_max
 from oml.utils.misc_torch import elementwise_dist, pairwise_dist
 
-TMetricsDict = Dict[str, Dict[int, Union[float, torch.Tensor]]]
+TMetricsDict = Dict[str, Dict[Union[int, float], Union[float, torch.Tensor]]]
 
 
 def calc_retrieval_metrics(
@@ -19,6 +19,7 @@ def calc_retrieval_metrics(
     cmc_top_k: Tuple[int, ...] = (5,),
     precision_top_k: Tuple[int, ...] = (5,),
     map_top_k: Tuple[int, ...] = (5,),
+    fmr_vals: Tuple[float, ...] = (1.0e-3,),
     reduce: bool = True,
     check_dataset_validity: bool = False,
 ) -> TMetricsDict:
@@ -33,6 +34,7 @@ def calc_retrieval_metrics(
         cmc_top_k: Tuple of ``k`` values to calculate ``cmc@k`` (`Cumulative Matching Characteristic`)
         precision_top_k: Tuple of  ``k`` values to calculate ``precision@k``
         map_top_k: Tuple of ``k`` values to calculate ``map@k`` (`Mean Average Precision`)
+        fmr_vals: Tuple of ``fmr`` values to calculate ``fnmr@fmr`` (`False Non Match Rate at given False Match Rate`)
         reduce: If ``False`` return metrics for each query without averaging
         check_dataset_validity: Set ``True`` if you want to check that we have available answers in the gallery for
          each of the queries
@@ -106,6 +108,9 @@ def calc_retrieval_metrics(
         positions = torch.arange(1, k_map + 1).unsqueeze(0)
         mean_ap = torch.sum((correct_preds / positions) * gt_tops[:, :k_map], dim=1) / n_gt_matrix
         metrics["map"][k_map_show] = mean_ap
+
+    for fmr_val in fmr_vals:
+        metrics["fnmr@fmr"][fmr_val] = calc_fnmr_at_fmr(distances, mask_gt, fmr_val)
 
     if reduce:
         metrics = reduce_metrics(metrics)
@@ -207,6 +212,14 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
         return acc.mean()
     else:
         return acc
+
+
+def calc_fnmr_at_fmr(distances: torch.Tensor, mask_gt: torch.Tensor, fmr_val: float) -> torch.Tensor:
+    pos_dist = distances[mask_gt]
+    neg_dist = distances[~mask_gt]
+    threshold = np.nanquantile(neg_dist.numpy(), fmr_val, interpolation="midpoint")
+    fnmr_at_fmr = (pos_dist >= threshold).sum() / len(pos_dist)
+    return fnmr_at_fmr
 
 
 def validate_dataset(mask_gt: torch.Tensor, mask_to_ignore: torch.Tensor) -> None:
--- a/oml/functional/metrics.py
+++ b/oml/functional/metrics.py
@@ -9,7 +9,7 @@ from oml.losses.triplet import get_tri_ids_in_plain
 from oml.utils.misc import clip_max
 from oml.utils.misc_torch import elementwise_dist, pairwise_dist
 
-TMetricsDict = Dict[str, Dict[Union[int, float], Union[float, torch.Tensor]]]
+TMetricsDict = Dict[str, Dict[int, Union[float, torch.Tensor]]]
 
 
 def calc_retrieval_metrics(
@@ -19,7 +19,7 @@ def calc_retrieval_metrics(
     cmc_top_k: Tuple[int, ...] = (5,),
     precision_top_k: Tuple[int, ...] = (5,),
     map_top_k: Tuple[int, ...] = (5,),
-    fmr_vals: Tuple[float, ...] = (1.0e-3,),
+    fmr_vals: Tuple[int, ...] = (1,),
     reduce: bool = True,
     check_dataset_validity: bool = False,
 ) -> TMetricsDict:
@@ -109,8 +109,11 @@ def calc_retrieval_metrics(
         mean_ap = torch.sum((correct_preds / positions) * gt_tops[:, :k_map], dim=1) / n_gt_matrix
         metrics["map"][k_map_show] = mean_ap
 
-    for fmr_val in fmr_vals:
-        metrics["fnmr@fmr"][fmr_val] = calc_fnmr_at_fmr(distances, mask_gt, fmr_val)
+    if len(fmr_vals) > 0:
+        pos_dist, neg_dist = extract_pos_neg_dists(distances, mask_gt, mask_to_ignore)
+        metric_vals = calc_fnmr_at_fmr(pos_dist, neg_dist, fmr_vals)
+        for fmr_val, metric_val in zip(fmr_vals, metric_vals):
+            metrics["fnmr@fmr"][fmr_val] = metric_val
 
     if reduce:
         metrics = reduce_metrics(metrics)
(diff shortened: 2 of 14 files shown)


tests/test_oml/test_functional/test_metrics/test_retrieval_metrics.py
```diff
@@ -162,3 +187,14 @@ def compare_metrics(
             values_expected = metrics_expected[metric_name][k]
             values_calculated = metrics_calculated[metric_name][k]
             assert torch.all(torch.isclose(values_expected, values_calculated, atol=1e-4)), [metric_name, k]
+
+
+def test_calc_fnmr_at_fmr(distances: torch.Tensor, mask_gt: torch.Tensor) -> None:
+    fmr_val = 0.1
+    fnmr_at_fmr = calc_fnmr_at_fmr(distances, mask_gt, fmr_val)
+    # positive distances are 0 0 1 1 2 2 5 5 9 9
```
AlekseySh, line 195: I guess this can be moved (or copied) to the original metric function `calc_fnmr_at_fmr`

tests/test_oml/test_functional/test_metrics/test_cmc_metric_old.py
```diff
@@ -38,26 +38,27 @@ def cmc_score(
 
 TEST_DATA_SIMPLE = (
     # (distance_matrix, mask_gt,  topk, expected_value)
-    (torch.tensor([[1, 2], [2, 1]]), torch.tensor([[0, 1], [1, 0]]), 1, 0.0),
-    (torch.tensor([[0, 0.5], [0.0, 0.5]]), torch.tensor([[0, 1], [1, 0]]), 1, 0.5),
-    (torch.tensor([[0, 0.5], [0.0, 0.5]]), torch.tensor([[0, 1], [1, 0]]), 2, 1),
+    (torch.tensor([[1, 2], [2, 1]]), torch.tensor([[0, 1], [1, 0]], dtype=torch.bool), 1, 0.0),
```
AlekseySh, line 41: I think that changing to bool is not related to the current PR. Let's roll back this file.

oml/metrics/embeddings.py
```diff
@@ -66,6 +66,7 @@ def __init__(
         cmc_top_k: Tuple[int, ...] = (5,),
         precision_top_k: Tuple[int, ...] = (5,),
         map_top_k: Tuple[int, ...] = (5,),
+        fmr_vals: Tuple[float, ...] = (0.1,),
```
AlekseySh, line 69: would `fmr_quantiles` be a better name for this argument?

oml/metrics/embeddings.py
```diff
@@ -85,6 +86,7 @@ def __init__(
             cmc_top_k: Values of ``k`` to compute ``CMC@k`` metrics
             precision_top_k: Values of ``k`` to compute ``Precision@k`` metrics
             map_top_k: Values of ``k`` to compute ``MAP@k`` metrics
+            fmr_vals: Values of ``fmr`` to compute ``FNMR@FMR`` metrics
```
AlekseySh, line 89: let's mention that these values are quantiles here

oml/functional/metrics.py
```diff
@@ -209,6 +214,14 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
         return acc
 
 
+def calc_fnmr_at_fmr(distances: torch.Tensor, mask_gt: torch.Tensor, fmr_val: float) -> torch.Tensor:
```
AlekseySh, line 217: Please, add the doc string explaining the idea of the metric and how we calculate it.
You may add an example with exact numbers from tests.
It would be also good to provide a link to the metric description. (Please, check other docs to see how we format links)

oml/functional/metrics.py
```diff
@@ -107,6 +109,9 @@ def calc_retrieval_metrics(
         mean_ap = torch.sum((correct_preds / positions) * gt_tops[:, :k_map], dim=1) / n_gt_matrix
         metrics["map"][k_map_show] = mean_ap
 
+    for fmr_val in fmr_vals:
+        metrics["fnmr@fmr"][fmr_val] = calc_fnmr_at_fmr(distances, mask_gt, fmr_val)
```
AlekseySh, line 113: We have a problem here with the logic of applying `mask_to_ignore` (we want some of the samples in the gallery don't affect the final metric, for existing metrics it may be done via masking the corresponding distances with inf values, check the code)

For the proposed metric this approach does not help since the quantile function does not know that some of the samples in the gallery have to be ignored. But NumPy docs say:
```
Compute the qth quantile of the data along the specified axis, while ignoring nan values.
```
As a solution, I can suggest costing infs to nans inside your function, but perhaps you may suggest a better way...

@DaloroAT what do you think?

AlekseySh (changes requested):
Thank you for your contribution! I left some comments

tests/test_oml/test_functional/test_metrics/test_cmc_metric_old.py
dapladoc: There was an error in `test_metric_greater` because `cmc_score_count` calls `calc_retrieval_metrics` and `fmr_vals` was set to default value `(1.e-3, )`. This made `calc_retrieval_metrics` call `calc_fnmr_at_fmr`, which in its turn had to evaluate `distances[mask_gt]` and pytorch threw the error because `mask_gt` is not int or bool type. So I will roll back this file, but I'll set `fmr_vals` to `tuple()` in `cmc_score_count` to supress `fnmr@fmr` evaluation.

oml/metrics/embeddings.py
dapladoc: Yeah, agree. And I think that maybe it will be better to make it not a float number in range [0, 1], but an int number in range [0, 100], like a percentile?

oml/metrics/embeddings.py
dapladoc: It could be confusing for users. What quantiles we are talking about? A user just need to evaluate FNMR@FMR, and quantiles are just a tool to archive this goal. So, we literally set desired values of FMR, and by coincidence this values are the quantiles of negative distances.

oml/functional/metrics.py
dapladoc: I was trying to mimic other functions in that module. I'll add the docstring.

oml/metrics/embeddings.py
AlekseySh: I do agree with the idea of using [0, 100] range

oml/metrics/embeddings.py
AlekseySh: Hm, I got your point of view. Should we at least mention it in the docs?
> we literally set desired values of FMR, and by coincidence this values are the quantiles of negative distances.

oml/functional/metrics.py
AlekseySh: thank you

oml/functional/metrics.py
AlekseySh: @dapladoc Do you understand the problem? If not, please, ask us questions

dapladoc: Yes, I understand the problem. I'll take into account `mask_to_ignore` not to consider these elements for the metric evaluation.

oml/metrics/embeddings.py
dapladoc: What do you think about the following?
```
fmr_vals: Values of ``fmr`` (measured in per cents) to compute ``FNMR@FMR`` metrics for
```

AlekseySh: Hm, what do you think about a more detailed description?

Values of ``fmr`` (measured in percents) for which we compute the corresponding ``FNMR``. For example, if ``fmr_values = (20, 40)`` we will calculate ``FNMR@FMR=20`` and ``FNMR@FMR=40``.

examples/cars/configs/train_cars.yaml
```diff
@@ -37,6 +37,7 @@ metric_args:
   metrics_to_exclude_from_visualization: [cmc,]
   cmc_top_k: [1]
   map_top_k: [5]
+  fmr_vals: [0.5, 0.1, 0.01, 0.001]
```
AlekseySh, line 40: Could you please find the most frequently used values of these metrics in the literature? 
For example, for the metrics we have, these values are `[1, 5, 10]`

dapladoc: In face recognition papers the most common values are 0.01% and 0.0001%. For fingerprints recognition I met 0% and 1%. I think that FMR values are really depends on the domain: in security-related domains it should be as small as possible, in other domains it could be much bigger. I think that 1% or 5% as a default value will be a good choice.

AlekseySh: Got it. Then let's replace the values in configs with those that you think are the good default choice

dapladoc: I set just 1%, because for the other metrics mostly only one `k` was set.

AlekseySh: Yes, that's also fine

AlekseySh:
@dapladoc Thank you for the contribution. When PR is ready to review, please, ping me here

dapladoc:
> @dapladoc Thank you for the contribution. When PR is ready to review, please, ping me here

@AlekseySh ping. I've done everything we discussed and even a little more

* Refactor `calc_fnmr_at_fmr` such that it takes `pos_dist` and `neg_dist` as its input; `pos_dist` and `neg_dist` could be required for some other functions in the future, so they are evaluated in the body of `calc_retrieval_metrics`;
* `calc_fnmr_at_fmr` now uses `numpy.percentile` assuming that `fmr_val` is a percentage;
* Add docstring to `calc_fnmr_at_fmr` with links to FNMR and FMR metrics, and an example;
* Rollback `test_cmc_metric_old.py`, so that `mask_gt` has type float;
* Change `fmr_vals` description in docstring of `EmbeddingMetrics`;
* `mask_to_ignore` now is used to extract distances;
* Change default value of `fmr_vals` to `[1]` in config files in examples.

By the way, the new `vulture` checker checks everything in the project directory, including my `venv` which I keep there, and it spits a lot of stuff (-: I added `venv` to excluded dirs in its settings in `pyproject.toml`.

tests/test_oml/test_functional/test_metrics/test_retrieval_metrics.py
```diff
@@ -44,6 +45,42 @@ def naive_precision(positions: TPositions, k: int) -> torch.Tensor:
     return metric
 
 
+@pytest.fixture()
+def distances() -> torch.Tensor:
```
AlekseySh, line 49: There is a problem in the current design: `test_calc_fnmr_at_fmr` pretend that it can process some distances or masks, but the values inside it are hardcoded for some particular inputs. So, it's not consistent. 

One of the solutions: let's put (dist, gt, ignore, fmr_vals, exprected_metric) into one fixture. Then we pass it to the test function, which calls the function and checks if the expected metric and the actual one are the same. 

If you have something different in mind it's also fine if it solves the problem

oml/functional/metrics.py
```diff
@@ -19,6 +19,7 @@ def calc_retrieval_metrics(
     cmc_top_k: Tuple[int, ...] = (5,),
     precision_top_k: Tuple[int, ...] = (5,),
     map_top_k: Tuple[int, ...] = (5,),
+    fmr_vals: Tuple[int, ...] = (1,),
```
AlekseySh, line 22: let's add assert for the range of all possible values

oml/functional/metrics.py
```diff
@@ -209,6 +218,55 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
...
+    is equal to ``fmr_val``.
+
+    Args:
+        pos_dist: distances between samples from the same class
+        neg_dist: distances between samples from different classes
+        fmr_vals: Values of ``fmr`` (measured in percents) for which we compute the corresponding ``FNMR``.
+                  For example, if ``fmr_values`` is (20, 40) we will calculate ``FNMR@FMR=20`` and ``FNMR@FMR=40``
+
+    See:
+        https://en.wikipedia.org/wiki/Biometrics#Performance
```
AlekseySh, line 233: please, format links in this way: https://github.com/OML-Team/open-metric-learning/blob/7e8406daab3b48946c81d12b5ed85c7b7f91351b/oml/samplers/balance.py#L17

it's needed if we want to add this function to the documentation (I guess we gonna do it later, with other similar functions after the refactoring that @dapladoc proposed before)

oml/functional/metrics.py
```diff
@@ -209,6 +218,55 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
...
+
+
+def extract_pos_neg_dists(
+    distances: torch.Tensor, mask_gt: torch.Tensor, mask_to_ignore: Optional[torch.Tensor]
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    """
+    Extract distances between samples from the same class, and distances between samples
+    in different classes.
+
+    Args:
```
AlekseySh, line 254: please, add docs for the objects you return

AlekseySh (changes requested):

AlekseySh:
As for `vulture`, it helped me to remove several variables that were unused when I set up confidence = 50% :)
But it also produces a lot of false positive alarms at this threshold

dapladoc:
@AlekseySh
Changes:
1. Remove fixtures from test_retrieval_metrics. I coded the `pos_dist` and `neg_dist` there because it looks like there will be no other tests that will require them.
2. Add asserts on `fmr_vals`. I think that this piece of code should be refactored, in order to add some meaningful error messages like `ValueError(f'{arg_name} should be a positive integer, but got {top_k_value}')` and `ValueError(f'fmr_vals should be integers in range [0, 100], but got {fmr_vals}')`.
3. Add the functions returns in docstrings, and make the links like they are supposed to be in sphinx documentation.

AlekseySh (approved):
Looks good!

AlekseySh:
@dapladoc, thank you for your contribution! 

By the way, do you want to perform some refactoring related to retrieval metrics in a dedicated PR?
I've made the issue for that: https://github.com/OML-Team/open-metric-learning/issues/240

dapladoc:
> By the way, do you want to perform some refactoring related to retrieval metrics in a dedicated PR?

Yes, I'd love to.

PR #245, opened 2022-11-29
Refactoring/calc retrieval metrics

dapladoc:
Hi guys!

Here is a draft of the `calc_retrieval` metrics refactoring. I'm still thinking about docstrings, external links in them and formulas. I will appreciate your opinions. Am I on the right track?

What was done:
* introduce new functions `calc_cmc`, `calc_precision`, `calc_map`, `_check_if_nonempty_integers_and_positive`, `_clip_max_with_warning`.
* add tests for the functions above. For each function 3 tests were defined: test when top-k is a tuple with several elements, test when top-k contains just 1 element (I added this test because it helps to understand the precise value of top-k for which the test could fail), and test when top-k tuple has not valid parameters.
* refactor `calc_retrieval_metrics` such that it uses the functions above.


oml/functional/metrics.py
```diff
@@ -220,42 +216,209 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
         return acc
 
 
+def calc_cmc(gt_tops: torch.Tensor, top_k: Tuple[int, ...]) -> List[torch.Tensor]:
+    """
+    Function to compute Cumulative Matching Characteristics (CMC) for each sample.
```
AlekseySh, line 221: I don't see the reason of keeping `for each sample`, probably, we can remove it?

AlekseySh: the same for other functions

oml/functional/metrics.py
```diff
@@ -220,42 +216,209 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
...
+    for k in top_k:
+        cmc.append(torch.any(gt_tops[:, :k], dim=1).float())
+    return cmc
+
+
+def calc_precision(gt_tops: torch.Tensor, n_gt: torch.Tensor, top_k: Tuple[int, ...]) -> List[torch.Tensor]:
+    """
+    Function to compute Precision@k for each sample.
+
+    ``precision@k`` is a proportion of the ``top_k`` elements from the same class to total number of the elements
```
AlekseySh, line 249: I would say

``precision@k`` is a fraction of the relevant gallery instances among ``top_k`` instances closest to the corresponding queries.

oml/functional/metrics.py
```diff
@@ -220,42 +216,209 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
...
+    if len(fmr_vals) == 0:
+        raise ValueError(f"fmr_vals are expected have at least one value, but got {fmr_vals}")
+    if not all(0 <= f <= 100 for f in fmr_vals):
+        raise ValueError(f"fmr_vals are expected to be integers in range [0, 100] but got {fmr_vals}")
     thresholds = torch.from_numpy(np.percentile(neg_dist.cpu().numpy(), fmr_vals)).to(pos_dist)
     fnmr_at_fmr = (pos_dist[None, :] >= thresholds[:, None]).sum(axis=1) / len(pos_dist)
     return fnmr_at_fmr
 
 
+def _calc_n_gt(mask_gt: torch.Tensor) -> torch.Tensor:
```
AlekseySh, line 409: I think having this small procedure in a separate function does not help readability, I guess we should put it back

AlekseySh (commented):
Thank you for your draft! Here are some comments from my side (By the way, I have not checked the tests yet):

- Let's avoid saying **from the same class**, it's too specific and not always the case. It's more about relevant outputs or not relevant ones. I suggested another wording in the comments.

- Please, start all of the comments for arguments with a capital letter (we do it in the rest of the project) 

- In the docstring sometime you put the metric name into the quotes, sometimes not. Let's unify it. I suggest using quotes for metric names only inside the docstrings for the particular arguments

- Let's change the formatting of the tensors in our examples (see comments)

README.md
```diff
@@ -15,9 +15,7 @@
 <div align="left">
 
 OML is a PyTorch-based framework to train and validate the models producing high-quality embeddings.
-
-## FAQ
-
+\n## FAQ\n
```
AlekseySh, line 18: what is the idea of changing the readme in this PR? (by the way, if you want to change Readme, please, check an instruction in a contribution guide)

dapladoc: It was done unintentionally. I didn't notice when it happened. I will revert the changes in README.md.

tests/test_oml/test_functional/test_metrics/test_retrieval_metrics.py
```diff
@@ -165,15 +170,160 @@ def compare_metrics(
...
+@pytest.mark.parametrize("top_k", (tuple(), (0, -1), (0,), (1.5, 2), (1.0, 2.0)))
+def test_calc_cmc_check_params(top_k: Tuple[int, ...]) -> None:
+    with pytest.raises(ValueError):
+        gt_tops = torch.ones((10, 5), dtype=torch.bool)
+        calc_cmc(gt_tops, top_k)
+
+
+def test_calc_precision(exact_test_case: Tuple[torch.Tensor, torch.Tensor, Tuple[int, ...], TMetricsDict]) -> None:
+    mask_gt, gt_tops, top_k, metrics_expected = exact_test_case
+    n_gt = _calc_n_gt(mask_gt)
```
DaloroAT, line 249: if you will decide to move `_calc_n_gt` calculation inside the function, you can also use

```python
@pytest.mark.parametrize("metric_function, metric_name", [(calc_cmc, 'cmc'), (calc_precision, 'precision'), (calc_map, 'map')])
``` 

and use as assert
```python
assert torch.all(
            torch.isclose(metrics_calculated, metrics_expected[metric_name][k], atol=1.0e-4)
        ), f"{metric_name}@{k} expected: {metrics_expected[metric_name][k]}; evaluated: {metrics_calculated}."
```

DaloroAT:
For me, PR is good in terms of implementation. I added some comments about docstrings and refactoring of tests to reduce the amount of them from 9 to 3 with `parametrize`.

oml/functional/metrics.py
```diff
@@ -220,42 +216,209 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
...
 
 
+def calc_cmc(gt_tops: torch.Tensor, top_k: Tuple[int, ...]) -> List[torch.Tensor]:
+    """
+    Function to compute Cumulative Matching Characteristics (CMC) for each sample.
+
+    ``cmc@k`` is 1 if ``top_k`` distances contain an element from the same class, and 0 otherwise.
+
+    Args:
+        gt_tops: ground truth of the ``max(top_k)`` distances.
```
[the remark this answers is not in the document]
DaloroAT: Decided to describe `gt_tops`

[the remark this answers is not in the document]
AlekseySh: I think let's keep the function as it is but use my comment above for a doc string

oml/functional/metrics.py
```diff
@@ -220,42 +216,209 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
...
+        _n_gt = torch.min(n_gt, torch.tensor(k).unsqueeze(0))
+        precision.append(correct_preds[:, k - 1] / _n_gt)
+    return precision
+
+
+def calc_map(gt_tops: torch.Tensor, n_gt: torch.Tensor, top_k: Tuple[int, ...]) -> List[torch.Tensor]:
+    """
+    Function to compute Mean Average Precision (MAP) for each sample.
+
+    ``map@k`` is the mean value of the ``precision@1``, ``precision@2``, ..., ``precision@k``
```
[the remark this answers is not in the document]
dapladoc: Could you give me a link on some article on MAP that you describe? All the articles I saw describe $map@k$ as

$map@k = \frac{1}{n}\sum\limits_{i = 1}^{k} precision@i \times rel(i)$,

where $rel(i)$ is 1 if $i$-th sample from the gallery relevant to the query an 0 otherwise.

See 
https://chaitanyabelhekar.medium.com/recommender-system-metrics-clearly-explained-1f2ba6690216
https://www.kaggle.com/code/nandeshwar/mean-average-precision-map-k-metric-explained-code/notebook
https://neptune.ai/blog/recommender-systems-metrics
https://machinelearninginterview.com/topics/machine-learning/mapatk_evaluation_metric_for_ranking/
https://www.coursera.org/lecture/basic-recommender-systems/ranking-metrics-F1gLc

Following your example, the $map@6$ calculated by the formula from links above shoud be

```
map@6 = 1/3 *(1/1 + 2/3 + 3/3)
```

dapladoc:
I've tried to take into account all the notes above, and make necessary changes.

* Group test functions by means of `pytest.mark.parametrize` according to @DaloroAT hints
* Reformat tensors in the docstrings examples
* Change docstrings for the `calc_*` functions. But I still need some clarification on `calc_map` function
* Change `calc_*` signatures: remove `n_gt` and add `mask_gt`. For consistency, `calc_cmc` also takes `mask_gt` as its input, even though it doesn't use it. Otherwise for `calc_cmc` we need to make its own rule in `test_metrics`.

As of the last point in the list above, it doesn't look good for me. `gt_top` is derived from `mask_gt`, so maybe it will be better to change the `calc_*` signatures so that they look like

```python
def calc_cmc(mask_gt: Tensor, top_indexes: torch.Tensor, top_k: Tuple[int, ...]) -> Tensor:
    """
    Args:
        mask_gt: ...
        top_indexes: Tensor of indices of the smallest k elements from the query-to-gallery distance matrix.
        ...
    """
```

But then the following lines will be done in three functions:

```python
    ii_arange = torch.arange(query_sz).unsqueeze(-1).expand(query_sz, max_k)
    gt_tops = mask_gt[ii_arange, ii_top_k]
```

Of course these lines should be placed in a separate function say `_extract_gt_tops`, but the computation still will take some time.

What do you think, @DaloroAT, @AlekseySh?

oml/functional/metrics.py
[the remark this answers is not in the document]
DaloroAT: Misunderstanding appears in the definition of `precision@k` in our repo and other resources. 

Usually, people define precision as `precision(k) = cumsum(gt)(k) / k` and `recall(k) = cumsum(gt)(k) / n_gt`.

But both metrics are not ok (if we log it or compare it with other people), because they don't take into account the relation between `k` and `n`. If the model is ideal, `precision(k)` will decrease with increasing `k` (when `k > n`) and `recall(k)` is low with low `k` (when `k < n`). So logging or comparing those metrics with other people is meaningless.

[Wiki says](https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Precision_at_k), that we can limit the denominator by the available amount of `n` for `k`, replacing the denominator to `min(k, n_gt)`. So in terms of OML, `precision@k = cumsum(gt)(k) / min(k, n_gt)`. And if we define `precision@k` in the repo in such a way, the statement that `map@k = mean(precision@1 + precision@2 + ... + precision@k)` is not correct.

Formula with `rel(i)` factor is correct, but with `precision(k) = cumsum(gt)(k) / k`.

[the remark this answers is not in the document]
dapladoc: I see. So, as of `precision@k` you take into account that `n` could be less than `k`, but as of `map@k` you don't take it into account. It looks a little odd though.

DaloroAT:
Hey @dapladoc 

- I wrote comments about `map` and `precision` related to `map`.
- You don't need to pass `mask_gt` to `calc_cmc`. A week ago we added a function, which keeps only necessary arguments for the specific function. So you can use it
```python
from oml.utils.misc import remove_unused_kargs


def func(a, b, c):
    print(a, b, c)


kwargs = {"a": 10, "b": 20, "c": 30, "d": 40}
print(kwargs)
kwargs = remove_unused_kargs(kwargs, func)
print(kwargs)
func(**kwargs)
```

oml/functional/metrics.py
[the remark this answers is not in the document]
DaloroAT: @dapladoc last notes about `map`

Probably, we should name our modified precision with another name. But in order not to create new names, we decided to refer to the wiki with `precision@k`.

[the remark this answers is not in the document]
DaloroAT: Resolve it one more time, if it's ok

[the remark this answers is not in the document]
dapladoc: Now it's completely ok. I understand the changes in the `precision` you've made and why you haven't made the same changes to the `map`, and this sounds very reasonable. The only problem is with the namings. I think this difference between your version of precision and the common one should be somehow describe in the docstrings.

I was using `MRR@k` and `Precision@k` (I used the version just like yours) in my project to track the position of the first relevant element among top `k` elements, and the precision.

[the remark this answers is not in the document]
AlekseySh: @dapladoc > I think this difference between your version of precision and the common one should be somehow describe in the docstrings.

Could you do this, please?

oml/functional/metrics.py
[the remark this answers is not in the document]
dapladoc: @DaloroAT, @AlekseySh

Actually, I suggested the changes in the `calc_*` below

> `gt_top` is derived from `mask_gt`, so maybe it will be better to change the `calc_*` signatures so that they look like

```python
def calc_cmc(mask_gt: Tensor, top_indexes: torch.Tensor, top_k: Tuple[int, ...]) -> Tensor:
    """
    Args:
        mask_gt: ...
        top_indexes: Tensor of indices of the smallest k elements from the query-to-gallery distance matrix.
        ...
    """
```

> But then the following lines will be done in three functions:

```python
    ii_arange = torch.arange(query_sz).unsqueeze(-1).expand(query_sz, max_k)
    gt_tops = mask_gt[ii_arange, ii_top_k]
```

> Of course these lines should be placed in a separate function say `_extract_gt_tops`, but the computation still will take some time.

What do you think about it? Is it ok to get `gt_tops` inside each `calc_*` function? Maybe some sort of a caching could be used here to reuse `gt_tops`, say simple `lru_cache` with `maxsize=1`. (I don't know whether it is safe to use `torch.Tensor` as a hashable variable, are there any drawbacks of this idea?)

If so, maybe we can also cache this `torch.topk(distances, k=max_k, largest=False)`, and change the `calc_*` signature to

```python
def calc_cmc(mask_gt: Tensor, distances: torch.Tensor, top_k: Tuple[int, ...]) -> Tensor:
```

oml/functional/metrics.py
```diff
@@ -220,37 +216,238 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
...
+
+    .. math::
+        \\begin{split}\\textrm{map}@k &=
+        \\frac{1}{\\min{\\left(k, n\\right)}}\\sum\\limits_{i = 1}^k
+        \\frac{\\textrm{# of relevant elements among top } i\\textrm{ elements}}{i} \\times \\textrm{rel}(i) = \\\\
+        & = \\frac{1}{\\min{\\left(k, n\\right)}}\\sum\\limits_{i = 1}^k
+        \\frac{\\sum\\limits_{j = 1}^{i}g_i}{i} \\times \\textrm{rel}(i)
+        \\end{split}
+
+    where :math:`\\textrm{rel}(i)` is 1 if :math:`i`-th element from the top :math:`i` clothest
```
AlekseySh, line 344: typo in clothest

oml/functional/metrics.py
```diff
@@ -220,37 +216,238 @@ def calculate_accuracy_on_triplets(embeddings: torch.Tensor, reduce_mean: bool =
...
-        pos_dist: 0 0 1 1 2 2 5 5 9 9
-        neg_dist: 3 3 4 4 6 6 7 7 8 8
-        10 percentile of negative distances is 3
-    The number of positive distances that are greater or equal than 3 is 4, therefore FNMR@FMR(10%) is 4 / 10
+
+    Example:
+        >>> pos_dist = torch.tensor([0, 0, 1, 1, 2, 2, 5, 5, 9, 9])
+        >>> neg_dist = torch.tensor([3, 3, 4, 4, 6, 6, 7, 7, 8, 8])
+        >>> fmr_vals = (10, 50)
+        >>> calc_fnmr_at_fmr(pos_dist, neg_dist, fmr_vals)
```
AlekseySh, line 443: let's do it in one line 
`calc_fnmr_at_fmr(pos_dist, neg_dist, fmr_vals=(10, 50))`

docs/source/contents/metrics.rst
```diff
@@ -20,6 +20,14 @@ EmbeddingMetrics
     .. automethod:: visualize
     .. automethod:: get_plot_for_queries
 
-calc_retrieval_metrics
+Metrics
```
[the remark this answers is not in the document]
AlekseySh: so, we will have a flatten structure of metrics module

oml/functional/metrics.py
```diff
@@ -297,6 +494,38 @@ def _to_tensor(array: Union[np.ndarray, torch.Tensor]) -> torch.Tensor:
         raise TypeError("Wrong type")
 
 
+def _check_if_nonempty_integers_and_positive(seq: Sequence[int], name: str) -> None:
```
AlekseySh, line 497: let's move it to `oml.utils.misc` and remove the underscore from the name
what do you think?

oml/functional/metrics.py
```diff
@@ -297,6 +494,38 @@ def _to_tensor(array: Union[np.ndarray, torch.Tensor]) -> torch.Tensor:
...
+    Args:
+        seq: A sequence.
+        name: A name of the sequence in case of exception should be raised.
+
+    """
+    if not len(seq) > 0 or not all([isinstance(x, int) and (x > 0) for x in seq]):
+        raise ValueError(f"{name} is expected to be a tuple with positive integers, but got {seq}")
+
+
+def _clip_max_with_warning(arr: Tuple[int, ...], max_el: int) -> Tuple[int, ...]:
```
AlekseySh, line 510: let's move it to `oml.utils.misc` and remove the underscore from the name
what do you think?

AlekseySh (commented):
Here are some comments, but I have not checked tests yet

oml/functional/metrics.py
dapladoc: I don't know whether this function could be useful in some other places, except functional/metrics.py.

oml/functional/metrics.py
dapladoc: I don't know whether this function could be useful in some other places, except functional/metrics.py.

AlekseySh:
@dapladoc as for the functions signatures, I do agree with you -- the current version should be changed.
My suggestion is the following:
```
calc_precision(gt_tops, n_gts, top_k)
calc_map(gt_tops, n_gts, top_k)
calc_cmc(gt_tops, top_k)
```

Basically, you started with this approach. But then we asked you to change it: https://github.com/OML-Team/open-metric-learning/pull/245#discussion_r1035798526.  At that time we thought that it was for performance reasons, but now it's clear that your option was also better in terms of logic. Sorry about that, let's roll it back, it you have no concerns.

dapladoc:
@AlekseySh why don't you like the idea of evaluating `gt_tops` right in `calc_*` functions, with caching `gt_tops` between `calc_*` calls?

<details>
<summary>Code sample</summary>


```python
def calc_retrieval_metrics(...):
    ...
    max_top_k = max(max(top_k) for top_k in (cmc_top_k, precision_top_k, map_top_k))
    # cache the gt_tops that will be used by all the `calc_*` functions
    _calc_gt_tops(distances, mask_gt, (max_top_k)) # here gt_tops will be evaluated
    cmc = calc_cmc(distances, mask_gt, cmc_top_k)
    precision = calc_precision(distances, mask_gt, precision_top_k)
    ...


def calc_cmc(distances, mask_gt, top_k):
    # if gt_tops already was computed for the distances and mask_gt and big enough
    # (has more than max(top_k) columns) then gt_tops will be taken from cache,
    # otherwise it will be evaluated
    gt_tops = _calc_gt_tops(distances, mask_gt, top_k)
    ...


def _calc_gt_tops(distances, mask_gt, top_k):
    # check if this function was called for these distances and mask_gt
    # and if gt_tops from the last run has more columns than max(top_k)
    key = _calc_gt_tops.key
    value = _calc_gt_tops.value
    if value and (distance, mask_gt) in key and max(top_k) <= value[0].shape[1]:
        return value[0]
    _, ii_top_k = torch.topk(distances, k=max_k, largest=False)
    ii_arange = torch.arange(query_sz).unsqueeze(-1).expand(query_sz, max_k)
    gt_tops = mask_gt[ii_arange, ii_top_k]
    _calc_gt_tops.key = set((distances, mask_gt))
    _calc_gt_tops.value = [gt_tops]
    return gt_tops
# make a storage for keeping results between runs
_calc_gt_tops.key = set()
_calc_gt_tops.value = []
```

</details>

Thus, all the `calc_*` functions will be self-contained, they will require nothing pre-computed in a particular order, just primary ingredients like distances, and ground truth matrix.

AlekseySh:
@dapladoc because I don't like the additional caching logic for `_calc_gt_tops`. Imho, bringing this complexity does not worth the goal. A signature like `calc_precision(gt_tops, n_gts, top_k)` seems okay to me if all of the arguments are explained in detail.

oml/functional/metrics.py
AlekseySh: At least here: https://github.com/OML-Team/open-metric-learning/blob/b25860c9f6018f85153963f270b3ecdb43ffc08d/oml/metrics/embeddings.py#L250, https://github.com/OML-Team/open-metric-learning/blob/b25860c9f6018f85153963f270b3ecdb43ffc08d/oml/metrics/embeddings.py#L256,

oml/functional/metrics.py
AlekseySh: As for this one -- I agree, it's more specific.

oml/functional/metrics.py
dapladoc: Sorry, but I don't see where `_check_if_nonempty_integers_and_positive` could be used in `get_plot_for_queries`. I only see some conditions on storage keys. What variable should contain positive integers only? Or you want to check that `query_ids` is not empty and has some indices?

AlekseySh: I am talking about `n_queries` and `n_instances`

dapladoc:
Move `check_if_nonempty_positive_integers` to `oml.utils.misc`.
Change `calc_*` signatures: remove `mask_gt` and add `n_gt`.
Add a note on `precision` in OML.
Other docstrings changes.

oml/functional/metrics.py
```diff
@@ -28,14 +28,15 @@ def calc_retrieval_metrics(
...
                      as answers and must be ignored
-        cmc_top_k: Tuple of ``k`` values to calculate ``cmc@k`` (`Cumulative Matching Characteristic`)
-        precision_top_k: Tuple of  ``k`` values to calculate ``precision@k``
-        map_top_k: Tuple of ``k`` values to calculate ``map@k`` (`Mean Average Precision`)
-        fmr_vals: Values of ``fmr`` (measured in percents) for which we compute the corresponding ``FNMR``.
-                  For example, if ``fmr_values`` is (20, 40) we will calculate ``FNMR@FMR=20`` and ``FNMR@FMR=40``
+        cmc_top_k: Values of ``k`` to calculate ``cmc@k`` (`Cumulative Matching Characteristic`)
+        precision_top_k: Values of ``k`` to calculate ``precision@k``
+        map_top_k: Values of ``k`` to calculate ``map@k`` (`Mean Average Precision`)
+        fmr_vals: Values of ``fmr`` (measured in percents) to calculate ``fnmr@fmr`` (False Non Match Rate
```
AlekseySh, line 37: minor: for the metrics above you used italic, let's do the same for this one

AlekseySh (commented):
When I look at examples I am confused because we have 3 queries and 3 top-ks, the output of the function is also 3x3 and it's hard to understand if the number of elements in tensors is aligned with a number of queries or with a number of top-ks. Let's remove one of the top-ks, then for a non-square setup (3 queries and 2 top-ks), it would be easier to interpret the results.

AlekseySh:
The code and docs of the functions look good!

As for the tests, is it possible to refactor them to avoid duplicated code fragments? (I am talking about `exact_case`)

dapladoc:
* Changed the examples in the `calc_*` docstrings
* Removed duplicated code in `test_retrieval_metrics.py`
* Added italic for `FNMR@FMR` in `calc_retrieval_metrics` docstring
* `cmc_top_k`, `precision_top_k` etc were copied from docstring of `calc_retrieval_metrics` to docstring of `EmbeddingMetrics`.

AlekseySh (approved):
LGFM! The metrics page in the documentation looks like almost a tutorial about different metrics

DaloroAT (commented):
Last fix. I hope you both have solved comments during the weekends

Let's merge

DaloroAT (dismissed):

DaloroAT (changes requested):
fix index

dapladoc:
> fix index

Done.

AlekseySh (approved):
LGFM!

AlekseySh:
@dapladoc thank you for the contribution!

PR #250, opened 2022-12-01
Removed FMR calculation to avoid OOM for SOP

AlekseySh:


dapladoc:
@AlekseySh
I thought about some algorithm to find quantiles online, in a batch mode. I found that there is P^2 algorithm [(link)](https://www.cse.wustl.edu/~jain/papers/ftp/psqr.pdf) and here is a discussion about it [(link)](https://stats.stackexchange.com/questions/7959/algorithm-to-dynamically-monitor-quantiles).
Then I found more articles about quantile estimation: see references to this article [(link)](https://oda.oslomet.no/oda-xmlui/bitstream/handle/10642/7943/Quantile_Multiplicative_Cybernetics.pdf?sequence=1). It looks like all these algorithms work only element-by-element. So, no way to parallelize it. I'm not a specialist in this, though.
Maybe it could be done effectively in cython. I'll try to implement the P^2 (or some other algorithm) in cython.

Another solution is to approximate the quantile (and the fnmr@fmr) by means of histogram, that could be computed incrementally. I can try to estimate accuracy of this approach on some syntetic data.

At the moment I don't have other ideas, but I will think on it, and maybe something new will come up.

AlekseySh:
@dapladoc I like the idea of an approximation. I thought about an even simpler solution: we can consider only the part of pairwise distances if the matrix is too big. For example, randomly pick 10% or something like this.

AlekseySh:
It's still unclear what is the problem: the footprint of copying distances or memory overhead inside the quantile algorithm.

AlekseySh:
@dapladoc Let's move the discussion here: https://github.com/OML-Team/open-metric-learning/issues/251

PR #254, opened 2022-12-06
Add doctests

### LINKED ISSUE #241: Introduce a test to check code examples in functions' doc strings
We introduced some code examples in doc strings for our functions. For example, here: 

https://github.com/OML-Team/open-metric-learning/blob/main/oml/functional/metrics.py#L237 

https://github.com/OML-Team/open-metric-learning/blob/main/oml/functional/metrics.py#L314 

https://github.com/OML-Team/open-metric-learning/blob/main/oml/functional/metrics.py#L380 

https://github.com/OML-Team/open-metric-learning/blob/main/oml/functional/metrics.py#L461

Illumaria:
Closes #241.

The `doctest` is only enabled for `oml/functional/metrics.py` now because some other files contain `doctest`-compatible syntax but the code is definitely not designed to be executed in a test ([here](https://github.com/OML-Team/open-metric-learning/blob/main/oml/interfaces/metrics.py#L24)'s an example).


Illumaria:
> 1. Probably it's better to ignore some files instead of using a whitelist of files which we want to check

Of course. But I gotta note here that, judging by the [`pytest` documentation](https://docs.pytest.org/en/7.1.x/how-to/doctest.html), it utilizes the Python built-in [`doctest`](https://docs.python.org/3/library/doctest.html#module-doctest) module which doesn't seem to have an option to ignore files. In this case, is adding a `# doctest: +SKIP` comment a convenient way for skipping a doctest or do you want something more flexible?

> 2. Does it compare the output with the provided expected one?

Sure, it's the way it works. :)

AlekseySh:
@Illumaria thank you for the answer. I'm thinking about the following: what if instead of finding a way to skip some of the files, we make all the examples work (I mean, refactor them a bit)?

AlekseySh:
If you use `>>>` to format your code, PyCharm understands that it's code and highlights the syntax and also properly applies refactoring manipulations. 

If needed, we may consider getting rid of this way of formatting

Illumaria:
> If you use `>>>` to format your code, PyCharm understands that it's code and highlights the syntax and also properly applies refactoring manipulations.
> 
> If needed, we may consider getting rid of this way of formatting

Alright, so I've searched for ways to sort this out, and it seems there's nothing we can do but to only use `>>>` syntax for doctests. Unfortunately, assuming that we remove it for all formatting cases, I wasn't able to find a good replacement in Sphinx, so any hints on how to best refactor this would be much appreciated.

AlekseySh:
I think that doctests are more important and I would say we may sacrifice `>>>`

Yes, refactoring functions are not available from PyCharm anymore, but we can find the mistakes caused by a lack of refactoring functionality in thanks to doctests, right?

Illumaria:
> I would say we may sacrifice `>>>`

I get it, but how exactly would you prefer to do it? We can simply drop it, or we can replace it with something different (I've tried various Sphinx formatting but nothing seemed suitable). As a particular [example](https://github.com/OML-Team/open-metric-learning/blob/main/oml/interfaces/metrics.py#L24) from above, I can turn

```python
    """
    Method for preparing metrics to work: memory allocation, placeholder preparation, etc.
    Has to be called before the first call of

    >>> self.update_data()

    """
```

into

```python
    """
    Method for preparing metrics to work: memory allocation, placeholder preparation, etc.
    Has to be called before the first call of self.update_data()
    """
```
Is such refactoring OK?

---

As for
> we can find the mistakes caused by a lack of refactoring functionality in thanks to doctests, right?

Correct, if we run doctests against the above-mentioned example, it will fail.

AlekseySh:
Got it. The proposed refactoring for this particular example seems valid to me. (But also add quotes around the code, please)

AlekseySh:
@Illumaria But what will happen to less trivial examples like this: https://github.com/OML-Team/open-metric-learning/blob/main/oml/functional/metrics.py#L237 ?

Illumaria:
> @Illumaria But what will happen to less trivial examples like this: https://github.com/OML-Team/open-metric-learning/blob/main/oml/functional/metrics.py#L237 ?

This one is OK, because it is a valid doctest:
Снимок экрана 2022-12-21 234303

AlekseySh:
@Illumaria got it!

AlekseySh:
@Illumaria Hey, I hope you are doing well. Any plans to finish this?

Illumaria:
@AlekseySh Hi! I'm really sorry for the delay. 😞 Please take a look at the changes, I did the refactoring that we had discussed previously and enabled doctests for the whole `oml` directory (I think there's no need to run doctests against the project root).

AlekseySh:
@Illumaria That's totally fine. Thank you for your work, I'll take a look

AlekseySh (commented):
Hey, @Illumaria, I have a few comments:

1) Do we really need to change the requirements in this PR? Seems like a confusing reshuffling.
2) Please, check & fix the formatting in "Read the Docs". (To see it simply click on the last icon in the list of CI tests for this PR).
3) Could we also keep the line-by-line structure of the keys?

Illumaria:
> Hey, @Illumaria, I have a few comments:
> 
>     1. Do we really need to change the requirements in this PR? Seems like a confusing reshuffling.
> 
>     2. Please, check & fix the formatting in "Read the Docs". (To see it simply click on the last icon in the list of CI tests for this PR).
> 
>     3. Could we also keep the line-by-line structure of the keys?

1. The order of requirements was changed in #248, but before that the order had been alphabetical, which I personally find more convenient. If you are okay with what #248 has brought, I'll revert the ordering changes that I've made. However, the change of pytest version from `>=6.0` to `>=7.2` is still mandatory for the doctests to work properly.
2. Sure, will do.
3. Of course.

AlekseySh:
@Illumaria cool, thx! But I ask you to revert the order changes in the requirements to keep the git history as clean as possible. The pytest version can be changed, it's fine. After that, I guess we are ready to merge your PR.

AlekseySh:
@Illumaria I run the CI, but there are some problems with GitHub actions today. So, it may fail and it's not your fault.

AlekseySh:
@Illumaria thank you for your contribution! It will help us to keep doc strings up to date and avoid unnecessary pain in the future! :)

PR #255, opened 2022-12-06
Introduce topological metrics.

dapladoc:
Add calc_topological_metrics.
Add calc_main_components_percentage.
Add tests for new functions.
Fix old tests to work with new functions.

For now `calc_main_components_percentage` returns _percentage_ of main component, so if the original embeddings dim is 512, and this function returns 45, then the actual number of main components is $512\cdot 0.45$.

### DIFF
--- a/examples/cars/configs/train_cars.yaml
+++ b/examples/cars/configs/train_cars.yaml
@@ -38,7 +38,7 @@ metric_args:
   cmc_top_k: [1]
   map_top_k: [5]
   fmr_vals: [1]
-  explained_variance_ths: [0.5, 0.9, 0.99]
+  pfc_explained_variance_ths: [0.5, 0.9, 0.99]
   return_only_main_category: True
   visualize_only_main_category: True
 
--- a/examples/cars/configs/train_cars.yaml
+++ b/examples/cars/configs/train_cars.yaml
@@ -38,6 +38,7 @@ metric_args:
   cmc_top_k: [1]
   map_top_k: [5]
   fmr_vals: [1]
+  explained_variance_to_keep: [50, 90, 99]
   return_only_main_category: True
   visualize_only_main_category: True
 
--- a/examples/cars/configs/train_cars.yaml
+++ b/examples/cars/configs/train_cars.yaml
@@ -38,7 +38,7 @@ metric_args:
   cmc_top_k: [1]
   map_top_k: [5]
   fmr_vals: [1]
-  pfc_explained_variance_ths: [0.5, 0.9, 0.99]
+  pfc_variance: [0.5, 0.9, 0.99]
   return_only_main_category: True
   visualize_only_main_category: True
 
--- a/examples/cars/configs/train_cars.yaml
+++ b/examples/cars/configs/train_cars.yaml
@@ -38,7 +38,7 @@ metric_args:
   cmc_top_k: [1]
   map_top_k: [5]
   fmr_vals: [1]
-  explained_variance_to_keep: [50, 90, 99]
+  explained_variance_ths: [0.5, 0.9, 0.99]
   return_only_main_category: True
   visualize_only_main_category: True
 
--- a/oml/functional/metrics.py
+++ b/oml/functional/metrics.py
@@ -9,7 +9,7 @@ from oml.losses.triplet import get_tri_ids_in_plain
 from oml.utils.misc import check_if_nonempty_positive_integers, clip_max
 from oml.utils.misc_torch import elementwise_dist, pairwise_dist
 
-TMetricsDict = Dict[str, Dict[int, Union[float, torch.Tensor]]]
+TMetricsDict = Dict[str, Dict[Union[int, float], Union[float, torch.Tensor]]]
 
 
 def calc_retrieval_metrics(
@@ -115,23 +115,25 @@ def calc_retrieval_metrics(
     return metrics
 
 
-def calc_topological_metrics(embeddings: torch.Tensor, explained_variance_to_keep: Tuple[int, ...]) -> TMetricsDict:
+def calc_topological_metrics(embeddings: torch.Tensor, explained_variance_ths: Tuple[float, ...]) -> TMetricsDict:
     """
     Function to evaluate different topological metrics.
 
     Args:
         embeddings: Embeddings matrix with the shape of ``[n_embeddings, embeddings_dim]``.
-        explained_variance_to_keep: Values in range [0, 100]. Find the number of components such that the amount
+        explained_variance_ths: Values in range [0, 1]. Find the number of components such that the amount
                                     of variance that needs to be explained is greater than the percentage specified
-                                    by ``explained_variance_to_keep``.
+                                    by ``explained_variance_ths``.
+
     Returns:
         Metrics dictionary.
+
     """
     metrics: TMetricsDict = dict()
 
-    if explained_variance_to_keep:
-        main_components = calc_main_components_percentage(embeddings, explained_variance_to_keep)
-        metrics["main_components"] = dict(zip(explained_variance_to_keep, main_components))
+    if explained_variance_ths:
+        main_components = calc_pcf(embeddings, explained_variance_ths)
+        metrics["pcf"] = dict(zip(explained_variance_ths, main_components))
 
     return metrics
 
@@ -432,6 +434,7 @@ def calc_fnmr_at_fmr(pos_dist: torch.Tensor, neg_dist: torch.Tensor, fmr_vals: T
         neg_dist: Distances between non-relevant samples.
         fmr_vals: Values of ``fmr`` (measured in percents) to compute the corresponding ``fnmr``.
                   For example, if ``fmr_values`` is (20, 40) we will calculate ``fnmr@fmr=20`` and ``fnmr@fmr=40``
+
     Returns:
         Tensor of ``fnmr@fmr`` values.
 
@@ -492,26 +495,24 @@ def calc_fnmr_at_fmr(pos_dist: torch.Tensor, neg_dist: torch.Tensor, fmr_vals: T
     return fnmr_at_fmr
 
 
-def calc_main_components_percentage(
-    embeddings: torch.Tensor, explained_variance_to_keep: Tuple[int, ...]
-) -> List[torch.Tensor]:
+def calc_pcf(embeddings: torch.Tensor, explained_variance_ths: Tuple[float, ...]) -> List[torch.Tensor]:
     """
-    Function estimates the number of main components of embeddings using Principal Component Analysis.
-    The number of components is estimated as a percentage of the embeddings dimension.
+    Function estimates the Principal Components Fraction (PCF) of embeddings using Principal Component Analysis.
+    The number of components is estimated as a fraction of the embeddings dimension.
 
     Args:
         embeddings: Embeddings matrix with the shape of ``[n_embeddings, embeddings_dim]``.
-        explained_variance_to_keep: Values in range [0, 100]. Find the number of components such that the amount
-                                    of variance that needs to be explained is greater than the percentage specified
-                                    by ``explained_variance_to_keep``.
+        explained_variance_ths: Values in range [0, 1]. Find the number of components such that the amount
+                                    of variance that needs to be explained is greater than the fraction specified
+                                    by ``explained_variance_ths``.
     Returns:
-        List of linear dimensions  as percentages of the embeddings dimension.
+        List of linear dimensions as a fractions of the embeddings dimension.
 
     Let :math:`X` be a set of :math:`d` dimensional embeddings.
     Let :math:`\\lambda_1, \\ldots, \\lambda_d\\in\\mathbb{R}` be a set of eigenvalues
     of the covariance matrix of :math:`X` sorted in descending order.
     Then for a given value of desired explained variance :math:`r`,
-    the number of main components that explaines :math:`r\\cdot 100\\%%` variance is the largest integer
+    the number of principal components that explaines :math:`r\\cdot 100\\%%` variance is the largest integer
     :math:`n` such that
 
     .. math::
@@ -520,7 +521,7 @@ def calc_main_components_percentage(
     The function returns
 
     .. math::
-        \\frac{n}{d}\\cdot 100\\%%
+        \\frac{n}{d}
 
     See:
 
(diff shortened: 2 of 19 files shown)


AlekseySh:
@dapladoc thx for the contribution! I will check it a bit later

AlekseySh (commented):
Thank you for the PR; let's start:

1. I think our metric needs a name. I suggest trying to find a suitable name for it in the literature. Otherwise, I suggest: `MeaningfulComponentsFraction (MCF)`. Then we can name its argument as `mcf_variances`, so it seems similar to other metrics.

2. I think that `explained_variance_to_keep` should be a float number in (0, 1) range to be aligned with the way how it's measured in sklearn. (the only problem is we have a  float as a key in a dictionary)

3. I think that returned value can be named as `meaningful_components_fraction` and it should be a float number in (0, 1). Now it's an integer, thus one may interpret 45 as 45 components left rather than 0.45 components are meaningful.

tests/test_oml/test_functional/test_metrics/test_topological_metrics.py
```diff
@@ -0,0 +1,59 @@
+from typing import Callable, Tuple
```
AlekseySh, line 1: After you have PCA as a separate function, you can test that it outputs the same as sklearn implementation. For this, I suggest you create a matrix of some values and copy the answer from sklearn here in order to use it as expected values.

To construct the matrix I suggest using `arrange()` in order not to deal with random and oversimplified examples constructed via `zeros` and `ones`. Please, check the alignment of intermediate values as well (not only the resulting matrix but explained variance as well).

oml/functional/metrics.py
```diff
@@ -465,15 +486,73 @@ def calc_fnmr_at_fmr(pos_dist: torch.Tensor, neg_dist: torch.Tensor, fmr_vals: T
...
+        ...                            [1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+        ...                            [0, 1, 0, 0, 0, 0, 0, 0, 0, 0],
+        ...                            [0, 0, 1, 0, 0, 0, 0, 0, 0, 0],
+        ... ], dtype=torch.float)
+        >>> calc_main_components_percentage(embeddings, explained_variance_to_keep=(50, 100))
+        tensor([10., 40.])
+
+    """
+    # The code below mirrors code from scikit-learn repository:
+    # https://github.com/scikit-learn/scikit-learn/blob/f3f51f9b6/sklearn/decomposition/_pca.py#L491
```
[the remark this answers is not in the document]
dapladoc: It looks like this should be not a function but a class. If it will be a function, it will return, for instance, `mean`, `singular_values`, `components` etc. But in future, to apply all this intermediate results for new vectors one has to know what should be done with all these values...

So I'd suggest to implement a simple class `PCA` like this

<details>
<summary><tt>PCA</tt></summary>

```python

class PCA:
    components: torch.Tensor
    mean: torch.Tensor
    singular_values: torch.Tensor
    explained_variance: torch.Tensor
    explained_variance_ratio: torch.Tensor

    def __init__(self, embeddings: torch.Tensor):
        self._fit(embeddings)
    
    def _fit(self, embeddings: torch.Tensor):
        ...
        self.mean = embeddings.mean(dim=0).unsqueeze(1)
        self.components = ...
        ...

    def transform(self, embeddings: torch.Tensor):
        ...
        embeddings = embeddings - self.mean
        ...
```
</details>


Afterwards all the `mean`, `explained_variance` etc. could be used as it is required to achieve some goals. For instance in my case I will be using only `explained_variance_ratio`.

dapladoc:
> 1. I think our metric needs a name. I suggest trying to find a suitable name for it in the literature. Otherwise, I suggest: `MeaningfulComponentsFraction (MCF)`. Then we can name its argument as `mcf_variances`, so it seems similar to other metrics.

There is nothing meaningful in these components. You can apply this function to random vectors, and it will be still random. So I can't agree with this name. In the sklearn's doc we can find the following

<details>
<summary>sklearn's docstring</summary>

```
 components_ : ndarray of shape (n_components, n_features)
        Principal axes in feature space, representing the directions of
        maximum variance in the data. Equivalently, the right singular
        vectors of the centered input data, parallel to its eigenvectors.
        The components are sorted by ``explained_variance_``.

n_components_ : int
        The estimated number of components. When n_components is set
        to 'mle' or a number between 0 and 1 (with svd_solver == 'full') this
        number is estimated from input data. Otherwise it equals the parameter
        n_components, or the lesser value of n_features and n_samples
        if n_components is None.
```

</details>

So I would suggest `MainComponentsFraction`, or `PrincipalComponentsFraction`. The latter I like more.

> 2\. think that `explained_variance_to_keep` should be a float number in (0, 1) range to be aligned with the way how it's measured in sklearn. (the only problem is we have a  float as a key in a dictionary). If we have the proposed metric name, we can simplify the name of this argument as well. For example: `variances_ths`

As of the name, I'd suggest `explained_variances_ths`.

If we are ok to have floats as a dict keys, then I would suggest to refactor `fnmr@fmr` to have values from [0, 1] there too.

What about converting the float numbers to `str` in those dicts?

```python
metric["main_componnts"][str(variance)] = 
```

Or using strings in the config files

```yaml
metric_args:
  explained_variances_ths: "0.5,0.9,0.99"
```

oml/functional/metrics.py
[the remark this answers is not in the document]
AlekseySh: @dapladoc class is even better

AlekseySh:
1. Okay, let's pick `PrincipalComponentsFraction`
2. `explained_variances_ths` is okay.

AlekseySh:
Af for `int` vs `float`. We discussed the problem with @DaloroAT. He pointed out that we have one more problem with ints: you said that for some of the tasks there is a big difference between `fmr=99` and `fmr=99.99`. With integers, we cannot express such a difference. So, seems like we need to switch to the float numbers, despite we will have them as keys. 

As for using string, I understand, that you want to keep a precision, but I hope we will have no problems with that since it makes no sense to pass more than 3 or 4 figures after the dot. So, let's try using a float, but please, run the experiment and check that lightning handles it correctly.

dapladoc:
* Add PCA class and tests for it
* Rename metric to `pcf` (principal components fraction)
* Test it with lightning, it works fine with float numbers
* Docstrings changes

oml/functional/metrics.py
```diff
@@ -465,15 +489,213 @@ def calc_fnmr_at_fmr(pos_dist: torch.Tensor, neg_dist: torch.Tensor, fmr_vals: T
...
+    _check_if_in_range(fmr_vals, 0, 100, "fmr_vals")
     thresholds = torch.from_numpy(np.percentile(neg_dist.cpu().numpy(), fmr_vals)).to(pos_dist)
     fnmr_at_fmr = 100 * (pos_dist[None, :] >= thresholds[:, None]).sum(axis=1) / len(pos_dist)
     return fnmr_at_fmr
 
 
+def calc_pcf(embeddings: torch.Tensor, explained_variance_ths: Tuple[float, ...]) -> List[torch.Tensor]:
+    """
+    Function estimates the Principal Components Fraction (PCF) of embeddings using Principal Component Analysis.
+    The number of components is estimated as a fraction of the embeddings dimension.
```
AlekseySh, line 501: I suggest the replacement:

> The number of components is estimated as a fraction of the embeddings dimension.

to 

> The metric is defined as a fraction of components needed to explain the required variance in data.

oml/functional/metrics.py
```diff
@@ -533,9 +755,26 @@ def _clip_max_with_warning(arr: Tuple[int, ...], max_el: int) -> Tuple[int, ...]
...
 
+def _check_if_in_range(vals: Sequence[float], min_: float, max_: float, name: str) -> None:
+    """
+    Check whether the ``vals`` are in the range ``[min_, max_]``. Throw the ValueError if not.
+
+    Args:
+        vals: Sequence to check.
+        min_: Minimal value of the range.
+        max_: Maximal value of the range.
+        name: name of the variable to throw the ValueError for.
```
AlekseySh, line 766: name -> Name

oml/metrics/embeddings.py
```diff
@@ -188,9 +195,12 @@ def compute_metrics(self) -> TMetricsDict_ByLabels:  # type: ignore
             mask_to_ignore=self.mask_to_ignore,
             check_dataset_validity=self.check_dataset_validity,
             reduce=False,
-            **args,  # type: ignore
+            **args_retrieval_metrics,  # type: ignore
         )
 
+        embeddings = self.acc.storage[self.embeddings_key]
+        metrics[self.overall_categories_key].update(calc_topological_metrics(embeddings, **args_topological_metrics))
```
AlekseySh, line 202: let's replace update with `=` by the analogue of retrieval metrics

oml/metrics/embeddings.py
```diff
@@ -204,9 +214,12 @@ def compute_metrics(self) -> TMetricsDict_ByLabels:  # type: ignore
                     mask_gt=self.mask_gt[mask],  # type: ignore
                     mask_to_ignore=self.mask_to_ignore[mask],  # type: ignore
                     reduce=False,
-                    **args,  # type: ignore
+                    **args_retrieval_metrics,  # type: ignore
                 )
 
+                mask = categories == category
+                metrics[category].update(calc_topological_metrics(embeddings[mask], **args_topological_metrics))
```
AlekseySh, line 221: the same ^

oml/functional/metrics.py
```diff
@@ -465,15 +489,213 @@ def calc_fnmr_at_fmr(pos_dist: torch.Tensor, neg_dist: torch.Tensor, fmr_vals: T
...
+    ratio_cumsum = torch.cumsum(pca.explained_variance_ratio, dim=0)
+    n_components = (
+        torch.searchsorted(ratio_cumsum, torch.tensor(explained_variance_ths).to(embeddings), side="right") + 1
+    )
+    return n_components / embeddings.shape[1]
+
+
+class PCA:
+    """
+    Principal component analysis (PCA).
```
AlekseySh, line 558: Let's add our motivation for having this: Note, the code is almost the same as one from sklearn, but we had two reasons to have our own implementation. First, we need to work with Torch tensors instead of NumPy arrays. Second, we wanted to avoid one more external dependency.

AlekseySh (changes requested):
PR looks solid, but I have some concerns related to PCA.
1) the place where PCA is implemented 
2) the place where PCA is tested 
3) the place in docs where PCA should be placed (I believe we should add it to docs, at least to have a possibility to check that big doc string that you've made)

tests/test_oml/test_functional/test_metrics/test_topological_metrics.py
```diff
@@ -0,0 +1,124 @@
+from typing import Callable, Dict, Tuple
+
+import pytest
+import torch
+
+from oml.functional.metrics import PCA, TMetricsDict, calc_pcf, calc_topological_metrics
+from oml.utils.misc import remove_unused_kargs
+
+
+def compare_dicts_recursively(d1: Dict, d2: Dict) -> bool:  # type: ignore
```
AlekseySh, line 10: By the way, why don't you use `d1 == d2`? Python is able to compare dicts recursively. The popular reason to do it manually is to check that values are close (`torch.close`), but not the same. Seems like it's not your case

oml/metrics/embeddings.py
dapladoc: Do you mean we need to add additional keys to distinguish the retrieval metrics and topological? Like the following

```python
metrics[self.overall_categories_key]['retrieval'] = calc_retrieval_metrics(...)
metrics[self.overall_categories_key]['topological'] = calc_topological_metrics(...)
```

tests/test_oml/test_functional/test_metrics/test_topological_metrics.py
dapladoc: I know, but I want to know exactly which keys differs. There is very similar code [here](https://github.com/OML-Team/open-metric-learning/blob/aaf6142c8bf737b576cc0c8cf43099bf9f62e39b/tests/test_oml/test_metrics/test_embedding_metrics.py#L24) but as for me it doesn't help much when debugging is required.

I don't like the duplicated code, but I don't want to touch that test module either. If you think it is ok, I can replace the code in `test_embedding_metrics.py` by this one. Where is it better to put this common function then?

oml/metrics/embeddings.py
AlekseySh: please, ignore this comment, I misunderstood what was going on there

tests/test_oml/test_functional/test_metrics/test_topological_metrics.py
AlekseySh: Okay, I see your point and agree. 
Then, put the function to utils and use it in both cases: your current tests and `test_embedding_metrics.py`

dapladoc:
* Move PCA from oml/functional/metrics.py to oml/utils/misc_torch.py.
* Rename `explained_variance_ths` to `pcf_explained_variance_ths`.
* Add calc_principal_axes_number to PCA.
* Move compare_dicts_recursively to oml/utils/misc.py and use it in tests.
* Docstrings changes.
* Add PCA to docs/source/contents/utils.rst.

tests/test_oml/test_utils/test_misc_torch.py
```diff
@@ -12,3 +12,68 @@ def test_elementwise_dist() -> None:
...
+def test_pca_inverse_transform() -> None:
+    embeddings = torch.eye(7, 6, dtype=torch.float)
+    embeddings = torch.cat((embeddings, embeddings), dim=1)
+    pca = PCA(embeddings)
+    embeddings_t = pca.transform(embeddings)
+    embeddings_it = pca.inverse_transform(embeddings_t)
+    assert torch.all(torch.isclose(embeddings, embeddings_it, atol=1.0e-6))
+
+
+def test_components_orthogonality() -> None:
```
AlekseySh, line 73: test_components_orthogonality -> test_pca_components_orthogonality

AlekseySh:
> There is docstrings duplication: the docstring for the PCA.calc_principal_axes_number() is almost completely the same as for 'calc_pcf'. Don't know what can be done to avoid it.

To avoid this duplication I suggest not having `calc_principal_axes_number` in PCA because it's more related to the metric. There is no analogue of this method in PCA from sklearn as well.

examples/cars/configs/train_cars.yaml
```diff
@@ -38,6 +38,7 @@ metric_args:
   cmc_top_k: [1]
   map_top_k: [5]
   fmr_vals: [1]
+  explained_variance_ths: [0.5, 0.9, 0.99]
```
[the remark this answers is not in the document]
AlekseySh: @dapladoc my idea was to use `pfc_variances` because `pfc_explained_variance_ths` is too long as I mentioned before. Anyway, pfc is our own metric, so, there is no chance for the argument to be self-explanatory, so, let's keep the short version: `pfc_variances`.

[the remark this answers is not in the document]
dapladoc: Sorry, I mislooked this name. I don't have objections against it.

dapladoc:
> To avoid this duplication I suggest not having `calc_principal_axes_number` in PCA because it's more related to the metric. Just put the whole logic into the metric function body. There is no analogue of this method in PCA from sklearn as well.

The analogue of this method in sklearn is built-in in the `PCA.fit`. Two reasons not to move `catl_principal_axes_number`:
1. AFAIK it is a kind of violation of best practices in OOP. If (or when) the `PCA` will be changed somehow, it is required to make changes in 2 places: `PCA` itself an `calc_pcf`. So, `calc_pcf` knows too much about PCA, and it increases code coupling.
2. In the future one could may have to estimate the number of components automatically to apply `PCA` in postprocessing. Thus, there should be some way to estimate and set the desired number of components automatically. If we move `calc_principal_axes_number` to pcf, then `PCA` will depend on it, and that increases coupling too.

dapladoc:
In the last commit `calc_principal_axes_number` is still a method in `PCA`. If you still think it is better to move it, I will move it.

AlekseySh:
@dapladoc As for PCA, I did not insist, I suggested it because you wanted to avoid the duplication in docs.

AlekseySh (approved):
LGFM! Thank you for your contribution!

PR #259, opened 2022-12-21
Postprocessing with Pairwise model

AlekseySh:


AlekseySh:
Changelog:
* Added `PairwiseEmbeddingsPostprocessor`
* Added `IPairwiseDistanceModel` and its implementation -- `SimpleSiamese` model
* Added `VectorsPairsDataset` and simple inference function for it 
* Added 2 helper functions: `take_2d`, `assign_2d` 
* Added tests for identity processing case

oml/postprocessors/pairwise_embeddings.py
```diff
@@ -0,0 +1,43 @@
...
+        # create (n_queries * top_n) pairs of each query and related galleries and re-estimate distances for them
+        emb_query = emb_query.repeat_interleave(top_n, dim=0)
+        emb_gallery = emb_gallery[ii_top]
+        distances_upd = pairwise_inference(self.model, emb_query, emb_gallery)
+        distances_upd = distances_upd.view(n_queries, top_n)
+
+        # update distances for top-n galleries, keeping the order of rest of the galleries (we use offset for it)
+        offset = distances_upd.min(dim=1)[0] - distances.min(dim=1)[0] + torch.finfo(torch.float32).eps
+        distances += offset.unsqueeze(-1)
+        distances = assign_2d(x=distances, indeces=ii_top.view(n_queries, top_n), new_values=distances_upd)
```
dapladoc, line 36: I'm not sure that it is a good idea to keep somehow old distances. It could be a source of very subtle errors, like when `top_n` here is less than say `cmc_top_k` or when a user just forgot that the postprocessor is turned on whereas he thinks that it is turned off. I'd suggest to spoil all the distances such that it will be clear that distance matrix has been changed somehow, and it will be a hint for a user to check what could have changed the distance matrix. I'd suggest to fill the rest of the distance matrix with `inf`.

dapladoc: And I don't know where it is better to add that we need to warn (or better to raise an error, because who reads warnings?) user that the `top_n` is less than `top_k` values for `cmc`, `precision` etc.

oml/metrics/embeddings.py
```diff
@@ -158,16 +159,19 @@ def _calc_matrices(self) -> None:
...
         self.mask_to_ignore = calc_mask_to_ignore(is_query=is_query, is_gallery=is_gallery)
         self.mask_gt = calc_gt_mask(labels=labels, is_query=is_query, is_gallery=is_gallery)
         self.distance_matrix = calc_distance_matrix(embeddings=embeddings, is_query=is_query, is_gallery=is_gallery)
 
+        if self.postprocessor:
+            self.distance_matrix = self.postprocessor.process(
+                distances=self.distance_matrix,
+                emb_query=embeddings[is_query],  # type: ignore
+                emb_gallery=embeddings[is_gallery],  # type: ignore
+            )
```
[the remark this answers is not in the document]
AlekseySh: Agree

oml/postprocessors/pairwise_embeddings.py
```diff
@@ -0,0 +1,43 @@
...
+
+from oml.datasets.vectors_pairs import pairwise_inference
+from oml.interfaces.models import IPairwiseDistanceModel
+from oml.interfaces.postprocessor import IPostprocessor
+from oml.utils.misc_torch import assign_2d
+
+
+class PairwiseEmbeddingsPostprocessor(IPostprocessor):
+    def __init__(self, pairwise_model: IPairwiseDistanceModel, top_n: int):
+        assert top_n > 0
```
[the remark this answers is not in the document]
AlekseySh: okay

oml/models/siamese.py
```diff
@@ -0,0 +1,27 @@
+import torch
+from torch import Tensor
+
+from oml.interfaces.models import IPairwiseDistanceModel
+from oml.utils.misc_torch import elementwise_dist
+
+
+class SimpleSiamese(IPairwiseDistanceModel):
```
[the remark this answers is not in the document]
AlekseySh: ok

oml/datasets/vectors_pairs.py
```diff
@@ -0,0 +1,45 @@
...
+        self.x2 = x2
+
+    def __getitem__(self, idx: int) -> Dict[str, Tensor]:
+        return {self.pair_1st_key: self.x1[idx], self.pair_2nd_key: self.x2[idx]}
+
+    def __len__(self) -> int:
+        return len(self.x1)
+
+
+def pairwise_inference(
```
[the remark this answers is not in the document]
AlekseySh: It's heavily related to the particular dataset, I suggest keeping it here for simplicity.

oml/datasets/vectors_pairs.py
```diff
@@ -0,0 +1,45 @@
...
+    model: IPairwiseDistanceModel, x1: Tensor, x2: Tensor, num_workers: int = 0, batch_size: int = 512
+) -> Tensor:
+    device = "gpu" if torch.cuda.is_available() else "cpu"
```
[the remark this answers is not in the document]
AlekseySh: agree

tests/test_oml/test_postprocessor/test_pairwise_embeddings.py
```diff
@@ -0,0 +1,64 @@
...
+    is_query = torch.ones(sz).bool()
+    is_gallery = torch.ones(sz).bool()
+
+    return embeddings, is_query, is_gallery
+
+
+@pytest.mark.parametrize("top_n", [1, 2, 5, 1000])
+@pytest.mark.parametrize("fixture_name", ["independent_query_gallery_case", "same_query_gallery_case"])
+def test_identity_processing(request: pytest.FixtureRequest, fixture_name: str, top_n: int) -> None:
+    embeddings, is_query, is_gallery = request.getfixturevalue("independent_query_gallery_case")
```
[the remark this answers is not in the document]
AlekseySh: oops, sure!

oml/postprocessors/pairwise_embeddings.py
```diff
@@ -0,0 +1,43 @@
+import torch
+from torch import Tensor
+
+from oml.datasets.vectors_pairs import pairwise_inference
+from oml.interfaces.models import IPairwiseDistanceModel
+from oml.interfaces.postprocessor import IPostprocessor
+from oml.utils.misc_torch import assign_2d
+
+
+class PairwiseEmbeddingsPostprocessor(IPostprocessor):
```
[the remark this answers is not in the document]
AlekseySh: okay

oml/postprocessors/pairwise_embeddings.py
AlekseySh: discussed offline

oml/utils/misc_torch.py
```diff
@@ -11,6 +11,51 @@
...
+    assert x.ndim == indeces.ndim == 2
+    assert x.shape[0] == indeces.shape[0]
+
+    n = x.shape[0]
+    ii = torch.arange(n).unsqueeze(-1).expand(n, indeces.shape[1])
+
+    return x[ii, indeces]
+
+
+def assign_2d(x: Tensor, indeces: Tensor, new_values: Tensor) -> Tensor:
```
DaloroAT, line 35: indices

oml/functional/metrics.py
```diff
@@ -153,10 +152,11 @@ def reduce_metrics(metrics_to_reduce: TMetricsDict) -> TMetricsDict:
 
 
 def apply_mask_to_ignore(
-    distances: torch.Tensor, mask_gt: torch.Tensor, mask_to_ignore: torch.Tensor
+    distances: torch.Tensor, mask_gt: Optional[torch.Tensor], mask_to_ignore: torch.Tensor
 ) -> Tuple[torch.Tensor, torch.Tensor]:
     distances[mask_to_ignore] = float("inf")
-    mask_gt[mask_to_ignore] = False
+    if mask_gt is not None:
```
DaloroAT, line 158: The such modification adds the possibility to make new errors in the future. And you use this modification to reduce 1 line of code in the new test. Worth it?

DaloroAT (changes requested):
don't use `list.py` name

oml/datasets/vectors_pairs.py
```diff
@@ -0,0 +1,54 @@
...
+        x2: Tensor,
+        num_workers: int = 0,
+        batch_size: int = 512,
+) -> Tensor:
+    dataset = VectorsPairsDataset(x1=x1, x2=x2)
+    loader = DataLoader(dataset, batch_size=batch_size, num_workers=num_workers, shuffle=False)
+
+    outputs = []
+    with torch.no_grad():
+        for batch in loader:
```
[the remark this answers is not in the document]
dapladoc: It could be very messy. For instance, if this function will be used somewhere in the middle of training step with its own progress bar, there will be a lot of bars. I'd suggest to add an argument like `verbose` for this purpose.

oml/functional/metrics.py
AlekseySh: ok

oml/utils/misc_torch.py
```diff
@@ -11,6 +11,51 @@
...
+
+    """
+    assert x.ndim == indices.ndim == new_values.ndim
+    assert x.shape[0] == indices.shape[0] == new_values.shape[0]
+    assert indices.shape == new_values.shape
+
+    n = x.shape[0]
+    ii = torch.arange(n).unsqueeze(-1).expand(n, indices.shape[1])
+
+    x[ii, indices] = new_values
```
dapladoc, line 54: Don't we need to be sure that `x` and `new_values` are on the same device?

oml/datasets/vectors_pairs.py
```diff
@@ -0,0 +1,61 @@
...
+    loader = tqdm(loader) if verbose else loader
+
+    outputs = []
+    with torch.no_grad():
+        for batch in loader:
+            x1 = batch[dataset.pair_1st_key].to(device)
+            x2 = batch[dataset.pair_2nd_key].to(device)
+            outputs.append(model(x1=x1, x2=x2))
+
+    return torch.cat(outputs).detach().cpu()
```
dapladoc, line 58: What about restoring `model.training` state in the end of the function?

oml/postprocessors/pairwise_embeddings.py
```diff
@@ -0,0 +1,80 @@
...
+        # 3. Update distances for top-n galleries
+        # The idea is that we somehow permute top-n galleries, but rest of the galleries
+        # we keep in the end of the list as before permutation.
+        # To do so, we add an offset to these galleries (which did not participate in permutation)
+        if top_n < n_galleries:
+            # Here we use the fact that distances not participating in permutation start with top_n + 1 position
+            min_in_old_distances = torch.topk(distances, k=top_n + 1, largest=False)[0][:, -1]
+            max_in_new_distances = distances_upd.max(dim=1)[0]
+            offset = max_in_new_distances - min_in_old_distances + 1e-5  # we also need some eps if max == min
+            distances += offset.unsqueeze(-1)
```
dapladoc, line 67: The logic is a little bit complicated. Isn't it enough to put `offset = max_in_new_distances + eps`? If the old distances were like 10, 20, 30, and the new one are 0.1, 0.2, 0.3? Then the offset is -9.7. Are we ok with shifting all the distances so much without necessity in this case? Let's make offset only a positive number in the end like `offset = max(1e-5, offset)`.

oml/utils/misc_torch.py
dapladoc: There will be an error if `x` and `values` are on different devices.

dapladoc (commented):
There are some issues with `pairwise_inference` and `assign2d` that are marked as resolved, but they still don't have required changes.

oml/datasets/vectors_pairs.py
dapladoc: It looks like it is the only option to guarantee that everything will work as expected. Otherwise someone can just forget to do `model.train()` after the `pairwise_distance` call.

dapladoc: Ok. I saw the comment above on it. We have different seeing of the library usage. It looks like you have in mind the config api usage, whereas I have in mind possibility to use parts of this library to build something new, and I wouldn't expect that function could change the `model.training` state and doesn't restore it. I accustomed that a function doesn't change the training flag, and it is my duty to do it, or it changes the flag, but restores it.

AlekseySh: Okay, I got your point

oml/datasets/vectors_pairs.py
```diff
@@ -0,0 +1,54 @@
...
+        x1: Tensor,
+        x2: Tensor,
+        num_workers: int = 0,
+        batch_size: int = 512,
+) -> Tensor:
+    dataset = VectorsPairsDataset(x1=x1, x2=x2)
+    loader = DataLoader(dataset, batch_size=batch_size, num_workers=num_workers, shuffle=False)
+
+    outputs = []
+    with torch.no_grad():
```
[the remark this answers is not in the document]
DaloroAT: But if you offer something, your users and developers should also follow it. If I want to reuse your function in another context, I will get some bugs.

oml/postprocessors/pairwise_embeddings.py
DaloroAT: I do not care about the exact value of the offset. It's only necessary not to get rounding errors, etc after modification.

DaloroAT (approved):

oml/postprocessors/pairwise_embeddings.py
AlekseySh: @dapladoc what would you say?

dapladoc: And what if we make it otherwise: not all the old distances shift to the new one, but the new one uplift to the old one? Thus, the property you described will be hold, and because `top_n` is expected to be small, we don't change original distribution so drastically. And this way we are getting along with the principle of least surprise: we claim that `top_n` queries will be re-ranked, and we change only distances for them.

AlekseySh: In this case, we need additional effort to avoid negative distance which may overcomplicate the logic (I mean we cannot subtract an offset anymore).

AlekseySh: PS. It's not clear what is the least surprising option in our case. In your approach, I cannot see that the outputs of the postprocessor have been written to the distance matrix, as I expect (because we changed them), right?

dapladoc: Yes, agree with you. It looks like the current implementation is better. Let's keep it.

PR #264, opened 2023-01-13
added reducer in criterion in PML example

IamSVP94:
but also need to change colab file


IamSVP94:
I return to that after reading contribution guide

AlekseySh:
@IamSVP94 thanks for your findings! 

Please, pay attention to how to contribute to Readme.md. It's described in the guide

PR #275, opened 2023-01-19
Added DDP inference code with handling duplicates occurred after DDP padding 

AlekseySh:


AlekseySh:
* Added DDP inference code with handling duplicates that occurred after DDP padding
* Added indices to getitem methods to keep track of duplicates

PR #280, opened 2023-01-22
Fix lru and don't force to have both train and val splits for ConfigAPI

DaloroAT:
- Now we don't require to have validation and train split to use our API.
- Added tests for checking dataframe format.
- If `cache_size == 0` or `cache_size is None` we don't use `lru_cache`. It allows us to launch ConfigAPI with `spawn` multiprocessing start method.

### DIFF
--- a/oml/datasets/base.py
+++ b/oml/datasets/base.py
@@ -50,7 +50,7 @@ class BaseDataset(Dataset):
         transform: Optional[TTransforms] = None,
         dataset_root: Optional[Union[str, Path]] = None,
         f_imread: TImReader = imread_cv2,
-        cache_size: int = 100_000,
+        cache_size: Optional[int] = 100_000,
         input_tensors_key: str = INPUT_TENSORS_KEY,
         labels_key: str = LABELS_KEY,
         paths_key: str = PATHS_KEY,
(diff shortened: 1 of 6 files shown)


oml/datasets/base.py
```diff
@@ -112,7 +112,9 @@ def __init__(
         self.df = df
         self.transform = transform if transform else get_transforms("norm_albu")
         self.f_imread = f_imread
-        self.read_bytes_image_cached = lru_cache(maxsize=cache_size)(self._read_bytes_image)
+        self.read_bytes_image_cached = (
```
AlekseySh, line 115: if you don't want to use cache you still have a function that has "cached" in the name 
let's rename it?

oml/datasets/list_dataset.py
```diff
@@ -48,7 +48,9 @@ def __init__(
         self.filenames_list = filenames_list
         self.transform = transform
         self.f_imread = f_imread
-        self.read_bytes_image_cached = lru_cache(maxsize=cache_size)(self._read_bytes_image)
+        self.read_bytes_image_cached = (
```
AlekseySh, line 51: the same ^

oml/datasets/triplet.py
```diff
@@ -59,7 +59,9 @@ def __init__(
         self.transforms = transforms or get_normalisation_albu()
         assert isinstance(transforms, albu.Compose) or (transforms is None)
 
-        self.read_bytes_image_cached = lru_cache(maxsize=cache_size)(self._read_bytes_image)
+        self.read_bytes_image_cached = (
```
AlekseySh, line 62: the same

AlekseySh (approved):
LGFM! Test covering in particular is really good.

I left only 1 small comment

DaloroAT:
`*_cached` removed

AlekseySh (approved):
Great!

PR #284, opened 2023-01-28
upd link to CUB datase

AlekseySh:


AlekseySh:
Have that link worked for you, @DaloroAT ?

DaloroAT:
Link is working, but we need to check converter and validation on the data

AlekseySh:
I've checked that converter works on data downloaded from this link. It worked.

PR #290, opened 2023-02-02
turn off fmr default metric for embeddings.py

Natyren:
and delete fmr_values in val_sop.yaml

For now, we don't push you to follow any predefined schema of issue, but ensure you've already read our contribution guide: https://open-metric-learning.readthedocs.io/en/latest/from_readme/contributing.html.

### DIFF
--- a/oml/metrics/embeddings.py
+++ b/oml/metrics/embeddings.py
@@ -67,7 +67,7 @@ class EmbeddingMetrics(IMetricVisualisable):
         cmc_top_k: Tuple[int, ...] = (5,),
         precision_top_k: Tuple[int, ...] = (5,),
         map_top_k: Tuple[int, ...] = (5,),
-        fmr_vals: Tuple[float, ...] = tuple(),
+        fmr_vals: Tuple[float, ...] = None,
         pfc_variance: Tuple[float, ...] = (0.5,),
         categories_key: Optional[str] = None,
         postprocessor: Optional[IDistancesPostprocessor] = None,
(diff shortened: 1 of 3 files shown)


AlekseySh:
@Natyren Oh, we already had an empty argument for the FNMR metric. I thought it was not empty. Thus, we don't need to change the code at all.

What we really need to do instead is to remove the metric from ``val_sop.yaml`` (already done in this PR) and from ``val_inshop.yaml`` (https://github.com/OML-Team/open-metric-learning/blob/main/examples/inshop/configs/val_inshop.yaml#L33)

Natyren:
@AlekseySh Change val_inshop.yaml, sorry for miscommunication on first commit

oml/metrics/embeddings.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         cmc_top_k: Tuple[int, ...] = (5,),
         precision_top_k: Tuple[int, ...] = (5,),
         map_top_k: Tuple[int, ...] = (5,),
-        fmr_vals: Tuple[float, ...] = tuple(),
+        fmr_vals: Tuple[float, ...] = None,
```
AlekseySh, line 70: we don't to change this line, it was already empty
so, roll back this particular line please

AlekseySh (changes requested):

oml/metrics/embeddings.py
Natyren: done

AlekseySh (approved):
LGFM, @Natyren thank you for your first contribution to OML!

PR #291, opened 2023-02-02
fix: updating last_log value in TripletLossPlain

alexmelekhin:
Fix for existing bug: `self.last_logs` value in `oml.losses.triplet.TripletLossPlain` is never updates and always returns empty dictionary which was set in `__init__` method.

### DIFF
--- a/oml/losses/triplet.py
+++ b/oml/losses/triplet.py
@@ -149,7 +149,10 @@ class TripletLossPlain(Module):
 
         anchor_ii, positive_ii, negative_ii = get_tri_ids_in_plain(n)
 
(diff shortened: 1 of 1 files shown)


AlekseySh:
@alexmelekhin Thanks, nice catch! We were not aware of this problem because this loss is not used in any pipeline since the implementation with a different signature is used instead. 

PS. How did you find that problem?

AlekseySh (approved):
Great!

alexmelekhin:
> PS. How did you find that problem?

I imported that loss directly and used it in my research.

PR #309, opened 2023-02-23
Postprocessing research

AlekseySh:


AlekseySh:
Changelog:
* removed `logs_images` and `logs_root` from all the validation configs because they have no effect (since we have no logger) 
* added dir related to embedded + postprocessor training

DaloroAT (dismissed):

examples/postprocessing/postprocessor_train.yaml
```diff
@@ -0,0 +1,124 @@
...
+      args:
+        normalise_features: False
+        use_multi_scale: False
+        weights: ${extractor_weights}
+        arch: vits16
+
+hard_pairs_mining: True
+
+optimizer:
+  name: adam
```
DaloroAT, line 54: adamw

examples/postprocessing/postprocessor_train.yaml
```diff
@@ -0,0 +1,124 @@
...
+  args:
+    lr: 2e-3
```
DaloroAT, line 56: 2e-3 - begining
1e-5 - rest

examples/postprocessing/postprocessor_validate.yaml
```diff
@@ -0,0 +1,66 @@
...
+    pairwise_model:
+      name: concat_siamese
+      args:
+        mlp_hidden_dims: [192]
+        weights: ${postprocessor_weights}
+        extractor:
+          name: vit
+          args:
+            arch: vits16
+            normalise_features: False
```
DaloroAT, line 41: false, true?

examples/postprocessing/visualize_postprocessing.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "        pairwise_model:\n",
+    "          name: concat_siamese\n",
+    "          args:\n",
+    "            mlp_hidden_dims: [192]\n",
+    "            weights: /path/to/postprocessor.ckpt\n",
+    "            extractor:\n",
+    "              name: vit\n",
+    "              args:\n",
+    "                arch: vits16\n",
+    "                normalise_features: False\n",
```
DaloroAT, line 100: true, false?

examples/postprocessing/postprocessor_train.yaml
AlekseySh: it's right, we have the scheduler later

examples/postprocessing/postprocessor_train.yaml
AlekseySh: oh, it's a fail
but I think it's better to keep the optimizer that we really used rather that mentioned in the paper 

and I also remember that there was no difference between adamw and adam when I trained an extractor

examples/postprocessing/visualize_postprocessing.ipynb
AlekseySh: False is correct, because it's inside the pairwise model

examples/postprocessing/postprocessor_validate.yaml
AlekseySh: False is correct, because it's inside the pairwise model

PR #312, opened 2023-03-02
Added ignoring of query-to-query interference before postprocessing 

AlekseySh:


oml/functional/metrics.py
```diff
@@ -16,7 +16,6 @@
 def calc_retrieval_metrics(
```
DaloroAT, line 16: There is no info about the structure of distances and mask_gt

DaloroAT: Also now we need to write 1 extra line with applying everywhere

oml/functional/metrics.py
```diff
@@ -545,20 +527,20 @@ def calc_pcf(embeddings: Tensor, pfc_variance: Tuple[float, ...]) -> List[Tensor
...
         distances: Distance matrix with the shape of ``[query_size, gallery_size]``
         mask_gt: ``(i,j)`` element indicates if for i-th query j-th gallery is the correct prediction
-        mask_to_ignore: Binary matrix to indicate that some elements in gallery cannot be used
-                     as answers and must be ignored
     Returns:
         pos_dist: Tensor of distances between relevant samples
         neg_dist: Tensor of distances between non-relevant samples
     """
+    mask_to_ignore = torch.isinf(distances)
     if mask_to_ignore is not None:
```
DaloroAT, line 544: Don't need to check it anymore

oml/functional/metrics.py
AlekseySh: what do you mean? we have a doc string

AlekseySh: > Also now we need to write 1 extra line with applying everywhere

it's okay because everywhere == 1 place in EmbeddingsMetrics, we don't have other entry points to the metrics in the library

oml/functional/metrics.py
AlekseySh: I want to have this check here in the case someone decides to call the function directly

AlekseySh:
@DaloroAT check my responses, please ^

oml/functional/metrics.py
DaloroAT: Screenshot from 2023-03-05 10-56-51
I mean now `mask_to_ignore` is `not None`, it's always `Tensor` (line 543).

oml/functional/metrics.py
DaloroAT: We have no info, that `apply_mask_to_ignore` have to be applied to `mask_gt` and `distances` before calling `calc_retrieval_metrics`.

Suppose, our potential users don't want to create `EmbeddingMetrics` with inner accumulator, keys, etc, if they have the necessary tensors and we offer a function with such good naming. 

We can also forget in several months, that `apply_mask_to_ignore` has to be called outside.

DaloroAT: Screenshot from 2023-03-05 11-06-59
We have ~10 places, where we call `calc_retrieval_metrics` and `apply_mask_to_ignore` should be added everywhere and we need don't forget to add it in the future.

DaloroAT: To be honest I don't really see why you decided to remove `mask_to_ignore` introducing potential bugs. You can add `apply_mask_to_ignore` before the postprocessor and don't change other lines. Calling `apply_mask_to_ignore` two times is extremely fast and doesn't produce errors.

AlekseySh: discussed offline

oml/functional/metrics.py
AlekseySh: discussed offline

DaloroAT (approved):
LADNO
Finally

yurikobyzev:
Hello,
Thank you for this grate framework!
I am study oml examples and find problem
with import "from oml.models.siamese import ConcatSiamese" which is not working in https://open-metric-learning.readthedocs.io/en/latest/examples/python.html 
Postprocessing example. 
Is it because there is another way of siamese realization in your framework?

AlekseySh:
@yurikobyzev thank you for using OML!

Based on the link you provided I see another code which imports the model:
`from oml.models.meta.siamese import ConcatSiamese` (this one is the newest one).
I also recommend you to update the library to the last version. 


PS. I'm a bit confused why you've decided to raise this problem here, in the closed issue. Don't hesitate to create a new one in the next time.

PR #316, opened 2023-03-05
Rework Zoo

AlekseySh:


oml/configs/model/resnet.yaml
```diff
@@ -1,7 +1,7 @@
 name: resnet
 args:
   arch: resnet50
-  gem_p: 7.0
+  gem_p: 5.0
```
DaloroAT, line 4: Any reason?

oml/models/resnet.py
```diff
@@ -31,11 +31,15 @@ class ResnetExtractor(IExtractor):
...
     pretrained_models = {
-        "resnet50_moco_v2": (
-            "https://dl.fbaipublicfiles.com/moco/moco_checkpoints/moco_v2_800ep/moco_v2_800ep_pretrain.pth.tar",
-            "a04e12f8",
-            None,
-        )
+        "resnet50_moco_v2": {
+            "url": "https://dl.fbaipublicfiles.com/moco/moco_checkpoints/moco_v2_800ep/moco_v2_800ep_pretrain.pth.tar",
+            "hash": "a04e12f8",
+            "fname": "moco_v2_800ep_pretrain.pth.tar",
```
DaloroAT, line 37: Does `fname` become mandatory? No `None` anymore?

oml/models/resnet.py
```diff
@@ -152,6 +160,26 @@ def draw_gradcam(self, image: np.ndarray) -> np.ndarray:
         img_with_grads = show_cam_on_image(image / 255, gray_image)
         return img_with_grads
 
+    @classmethod
+    def from_pretrained(cls, weights: str) -> "ResnetExtractor":
+        if weights.lower().endswith("_default") or weights.lower().endswith("_imagenet_v1"):
```
DaloroAT, line 165: Why do we need some custom code here instead of a declaration in `pretrained_models`?

In the case of ViT we have a lot of external checkpoints and they are in `pretrained_models`. But for some reason, it's not the case for resnet torchvision checkpoints.

oml/models/vit/clip.py
```diff
@@ -15,37 +15,65 @@
...
+            "normalise_features": False,
+            "fname": "openai_vitb16_224.ckpt",
+            "arch": "vitb16_224",
+        },
+        "openai_vitb32_224": {
+            "url": f"{_OPENAI_URL}/40d365715913c9da98579312b702a82c18be219cc2a73407c4526f58eba950af/ViT-B-32.pt",
+            "hash": "3ba34e387b24dfe590eeb1ae6a8a122b",
+            "is_jitted": True,
+            "normalise_features": False,
+            # "fname": "openai_vitb32_224.ckpt",
```
DaloroAT, line 31: comments

oml/models/resnet.py
DaloroAT: If you remove custom code, it will be possible to move `from_pretrained` to `Extractor`, read `pretrained_models`, remove keys with `remove_unused_kwargs` and pass them to `__init__`

oml/models/resnet.py
AlekseySh: It's not mandatory, but I think having `fname` is good because of 2 reasons:
1) names can overlap between different pretrained checkpoints, since the file name on a disk and the name in URL may be different 
2) because of the reason above, it's also easier to understand where is the ckpt file related to the pretrained ckpt that you downloaded

oml/models/vit/clip.py
AlekseySh: thx

oml/models/resnet.py
AlekseySh: It's a mess with ResNet, I know. I already created an issue to fight with that in the next PR. It will be easier to review then. After that, I can unify `from_pretrained` if possible (or, I will do it altogether with refactoring of `ResnetExtractor`)

oml/models/resnet.py
DaloroAT: Just  a question

oml/configs/model/resnet.yaml
DaloroAT: I mean checkpoint doesn't change but the parameter is changed. Ok

DaloroAT (approved):

PR #321, opened 2023-03-12
Added tests (transforms for pretrained and visualisations)

AlekseySh:


AlekseySh:
Changelog:
1. We test that all the pretrained extractors have the corresponding transforms 
2. We test functions that  help to interpret models' predictions

tests/test_oml/test_models/test_visualisations.py
```diff
@@ -0,0 +1,43 @@
...
+    ],
+)
+@pytest.mark.parametrize(
+    "image",
+    [
+        get_numpy_image(),
+        get_image_pillow(),
+    ],
+)
+def test_visualisation(draw_function: Any, image: Union[np.ndarray, Image.Image]) -> None:
```
dapladoc, line 35: Maybe it makes sense to assure that for the same image in `numpy` and `PIL` formats the result should be the same.

AlekseySh: I think it changes the logic of the test too much without a lof of additional value, because under the hood in the visualization functions everything is cast to numpy, so, there is nothing really to check

dapladoc: As far as I know, taking into account the things under the hood of a function is not a correct way to write tests in this case. The implementation of the function may change in the future. Or maybe there will be some issues with other versions of numpy or PIL, so that they will produce different results as outputs of the function.

Maybe we can add another function to make such test.

tests/test_oml/test_models/test_visualisations.py
```diff
@@ -0,0 +1,43 @@
...
+
+IMAGE_SIZE = (224, 224, 3)
+
+
+def get_image_pillow() -> Image.Image:
+    img = np.random.randint(0, 256, size=IMAGE_SIZE, dtype=np.uint8)
+    return Image.fromarray(img)
+
+
+def get_numpy_image() -> np.ndarray:
```
DaloroAT, line 17: If you decided to test several cases, it would be great to check different `dtype`s for input.

tests/test_oml/test_models/test_visualisations.py
AlekseySh: Basically, you are right, I was to lazy to accept it :) 

Well, I added one more test

tests/test_oml/test_models/test_visualisations.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.mark.parametrize(
+    "draw_function",
+    [
+        ViTExtractor(None, "vits16", normalise_features=False).draw_attention,
+        ResnetExtractor(None, "resnet50", normalise_features=False, gem_p=None, remove_fc=False).draw_gradcam,
+    ],
+)
+def test_visualisation_2(draw_function: Any) -> None:
```
dapladoc, line 53: Type annotation for `draw_function` could be more precise. Or are we ok with `Any` in this case?

AlekseySh: for tests it's okay

tests/test_oml/test_models/test_visualisations.py
AlekseySh: I don't really want to overcomplicate it now and support different formats. Instead, I've added a docstring where I describe that we want an image with pixel values in the range of ``[0..255]``.

oml/models/resnet.py
```diff
@@ -167,7 +173,7 @@ def draw_gradcam(self, image: Union[np.ndarray, TPILImage]) -> Union[np.ndarray,
         if self.remove_fc:
             raise ValueError("This method does not work if there is no FC layer in the model.")
 
-        need_to_convert = type(image) == TPILImage
+        need_to_convert = type(image) != np.ndarray
```
dapladoc, line 176: I'd suggest a more pythonic way

```python
need_to_convert = not isinstance(image, np.ndarray)
```

And the same for `vit`.

AlekseySh: okay

tests/test_oml/test_models/test_visualisations.py
DaloroAT: If you want [0..255], you could add assert for ranges or np.uint8

dapladoc (approved):

DaloroAT (approved):

PR #322, opened 2023-03-12
Fixed configs: models for CAR and CUB were trained with bboxes 

AlekseySh:


dapladoc (changes requested):
`examples/datasets_converters/README.md` needs to be changes. It should be mentioned that for CAR and CUB there should be no option ` --no_bboxes` in the argument list for correspondent converter.

I'd suggest to explicitly add the conversion step in the correspondent `README.md` file (i.e. `examples/extractor_cars/README.md`) in order to make the examples self-contained.

dapladoc (approved):

PR #325, opened 2023-03-13
Reworked constructors in pretrained models

AlekseySh:


oml/models/projection.py
```diff
@@ -75,25 +85,13 @@ def unfreeze(self) -> None:
...
-                }
-            }
-        )
+        # The current class takes another model as a constructor's argument, so, they need to be
+        # in the `self.pretrained_models`. The problem is these models will be instantiated even if we simply
+        # import something from the current module. To avoid it we added the logic of wrapping/unwrapping
+        # constructors into lambda functions.
+
+        ini = cls.pretrained_models[weights]["init_args"]
+        ini["extractor"] = ini["extractor"]()  # type: ignore
```
dapladoc, line 94: Wouldn't it be better/cleaner to separate `extractor` and a function that creates it? I mean something like

```python
pretrained_models = {
        "vits16_224_mlp_384_inshop": {  # type: ignore
            "url": f"{STORAGE_CKPTS}/inshop/vits16_224_mlp_384_inshop.ckpt",
            "hash": "35244966",
            "fname": "vits16_224_mlp_384_inshop.ckpt",
            "init_args": {
                "extractor_factory": lambda: ViTExtractor(None, "vits16", normalise_features=False, use_multi_scale=False),
                "mlp_features": [384],
                "train_backbone": True,
            },
        }
    }
...
ini = cls.pretrained_models[weights]["init_args"]
ini["extractor"] = ini["extractor_factory"]()
...
```

oml/models/projection.py
```diff
@@ -17,8 +17,18 @@ class ExtractorWithMLP(IExtractor, IFreezable):
...
 
-    # We update this dictionary later, see `self.from_pretrained()` method
-    pretrained_models: Dict[str, Any] = {"vits16_224_mlp_384_inshop": None}
+    pretrained_models = {
+        "vits16_224_mlp_384_inshop": {  # type: ignore
+            "url": f"{STORAGE_CKPTS}/inshop/vits16_224_mlp_384_inshop.ckpt",
+            "hash": "35244966",
+            "fname": "vits16_224_mlp_384_inshop.ckpt",
+            "init_args": {
+                "extractor": lambda: ViTExtractor(None, "vits16", normalise_features=False, use_multi_scale=False),
```
DaloroAT, line 26: Will it work during the train? Multiprocessing doesn't like lambdas very much.

DaloroAT:
It seems that this problem is not very flexible to be solved. The most obvious thing that can be done is to write functions (or methods) for each pretrained model to create it. Or do it only for edge cases.

In the `.pretrained_models`, have either `init_args` or a reference to a function (method) that will return an instance.

dapladoc (approved):

oml/models/projection.py
DaloroAT: Good

PR #326, opened 2023-03-13
Neptune warning

AlekseySh:


oml/lightning/entrypoints/parser.py
```diff
@@ -68,6 +69,12 @@ def initialize_logging(cfg: TCfg) -> Union[bool, NeptuneLogger]:
 
     """
     if ("NEPTUNE_API_TOKEN" in os.environ.keys()) and (cfg["neptune_project"] is not None):
+        warnings.warn(
+            "Unfortunately, in the case of using Neptune, you may experience that long experiments are"
+            "stacked and not responding. It's not an issue on OML's side, so, we cannot fix it. You can use"
+            "Tensorboard logger instead, for this simply leave <NEPTUNE_API_TOKEN> unfilled."
+        )
+
```
dapladoc, line 77: If a long experiment is not responding for a while, will it finally continue or is it stuck forever? If we know that at some point the experiment continues normally, we can include this information in the warning.

I stumbled upon these situations and each time after a while everything went well. In my case these situations were caused by the quality of my network connection (Neptune needs to upload all the data to its server and it took very long time to do it via my network).

dapladoc (approved):

PR #329, opened 2023-03-13
Added extra args for PL Trainer to Config-API

AlekseySh:


oml/lightning/entrypoints/train.py
```diff
@@ -134,6 +134,7 @@ def pl_train(cfg: TCfg) -> None:
         logger=logger,
         precision=cfg.get("precision", 32),
         **trainer_engine_params,
+        **cfg.get("extra_args_for_lightning_trainer", {}),
```
dapladoc, line 137: Following the other namings in Config-API I'd suggest to call it `lightning_trainer_extra_args`.

AlekseySh: okay

dapladoc (approved):

PR #330, opened 2023-03-14
Converter fix

AlekseySh:


examples/datasets_converters/convert_inshop.py
```diff
@@ -68,10 +68,10 @@ def build_inshop_df(dataset_root: Path, no_bboxes: bool) -> pd.DataFrame:
 
     cols_to_pick = ["label", "path", "split", "is_query", "is_gallery", "category"]
 
-    if no_bboxes:
-        df["path"] = df["image_name"].apply(lambda x: Path(dataset_root) / x.replace("img/", "img_highres/"))
+    df["path"] = df["image_name"].apply(lambda x: Path(dataset_root) / x.replace("img/", "img_highres/"))
 
-    else:
+    need_bboxes = not no_bboxes
```
dapladoc, line 73: Do you really need to introduce a new variable just to use it in one line below?

examples/datasets_converters/convert_inshop.py
```diff
@@ -80,7 +80,6 @@ def build_inshop_df(dataset_root: Path, no_bboxes: bool) -> pd.DataFrame:
 
         df = df.merge(df_bbox, on="image_name", how="inner")
         df.reset_index(inplace=True, drop=True)
-        df["path"] = df["image_name"].apply(lambda x: Path(dataset_root) / x)
```
dapladoc, line 83: The logic of `path` construction differs. I don't know the details, but is it ok that in the new version of the code we always do `x.replace("img/", "img_highres/")`?

examples/datasets_converters/convert_inshop.py
AlekseySh: for better readability

dapladoc: Can't agree with you that it is more readable. You have `no_bboxes` argument of the function that is used just once to create a new variable `need_bboxes`. As I see it, it is better to rename the argument `no_bboxes` to `need_bboxes` in the function signature, and possibly in the script arguments ([line 123](https://github.com/OML-Team/open-metric-learning/blob/556f692f83a669c57bbf60f2efccf55f350abcc5/examples/datasets_converters/convert_inshop.py#L123))

PR #350, opened 2023-04-02
Reworked Readme and Readthedocs

AlekseySh:


AlekseySh:
Changelog:
* Split python examples into separate files so they can be used independently
* Kept only two python examples in readme (train/val + usage)
* United documentation and extra materials
* Reworked structure of Readthedocs with respect to plan

ToDo:
Fix the existing links, and add the new ones among different subpages. It may be hard before this and another PR have been merged.

DaloroAT (approved):

PR #351, opened 2023-04-03
Added docs for Pipelines

AlekseySh:


pipelines/README.md
```diff
@@ -1,45 +1,134 @@
...
-We use [Hydra](https://hydra.cc/docs/intro/) as a parser for `.yaml` configs.
-So, you can change whatever you want directly in the config file or override some parameters
-using command line interface:
-```
-python train_cars.py optimizer.args.lr=0.000001 bs_val=128
+
+Pipelines are a predefined collection of scripts/recipes that provide a way to run metric learning
+experiments by changing only the config.
+You need to prepare the `.csv` file which describes your dataset
+(the format is explained [here](https://open-metric-learning.readthedocs.io/en/latest/oml/data.html)).
```
dapladoc, line 11: This sentence looks not related to the previous one. It is like out of nowhere.

Maybe we cat write
```
Having `.csv` file which describes your dataset pipelines are a predefined collection...
```

but it doesn't look much better.

DaloroAT:
Use `subprocess` to better handle exit codes

image

pipelines/README.md
```diff
@@ -1,45 +1,134 @@
...
+
+They will not work if:
+* You deal with a corner case and the flexibility of an existing Pipeline isn't enough
+
+## Minimal example of a Pipeline
+
+Each Pipeline is built around 3 components:
+* Config file
+* Registry of classes & functions
+* Entrypoint function for a Pipeline + script to run it
```
dapladoc, line 26: We can write about `.csv` file here.

pipelines/README.md
```diff
@@ -1,45 +1,134 @@
...
+python run.py model.args.weights=null
+
+```
+[comment]:shell-end
+
+Note, we use [Hydra](https://hydra.cc/docs/intro/) as a config parser. One of its abilities
+is to change part of the config from a command line, as showed above.
+
+
+## How to work with Pipelines?
```
dapladoc, line 101: I would move this section before `Minimal example of a Pipeline`.

pipelines/README.md
```diff
@@ -1,45 +1,134 @@
...
+
+The recommended way is the following:
+1. Install OML: `pip install open-metric-learning`
+2. Prepare your dataset in the required [format](https://open-metric-learning.readthedocs.io/en/latest/oml/data.html). (There are [converters](https://github.com/OML-Team/open-metric-learning/tree/main/pipelines/datasets_converters) for 4 popular datasets).
+3. Go to Pipeline's folder and copy `.py` script and its `.yaml` to your workdir. Modify the config if needed.
+4. Run the script via the command line.
+
+## Building blocks of Pipelines
```
dapladoc, line 109: I would move this section before `Minimal example of a Pipeline`.

pipelines/README.md
```diff
@@ -88,4 +179,33 @@ if __name__ == "__main__":
...
+
+Let's say you want to change the implementation of `Dataset`, which is not configurable
+in a pipeline of your interest. In other words, you can only change its initial arguments,
+but cannot replace the corresponding class.
+
+In this case, you can copy the source code of the main pipeline
+entrypoint function and modify it as you want.
+For example, if you want to train your feature extractor with your own implementation of `Dataset`,
+you need to copy & modify
+[extractor_training_pipeline](https://open-metric-learning.readthedocs.io/en/latest/contents/lightning.html#extractor-training-pipeline).
```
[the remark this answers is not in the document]
AlekseySh: okay

pipelines/README.md
AlekseySh: I made the transition smoother: `Pipelines require your data to pe prepared in a special `.csv` file, described in details here.`

pipelines/README.md
AlekseySh: We describe Pipeline here and its components. I don't think it's a good place for mentioning csv here.

pipelines/README.md
AlekseySh: Agree

pipelines/README.md
AlekseySh: I cannot agree here, because there is a transition from `Building blocks of Pipelines`  to `How to use my own implementation of loss, model, etc.?`. If I break it with a `Minimal example of a Pipeline` it would break the narrative.

PR #358, opened 2023-04-26
Removed ambiguous of using both "extractor" and "model" names for extractor

AlekseySh:


DaloroAT:
Check `.py` files when we use registry

https://github.com/OML-Team/open-metric-learning/blob/b0a020371080d6d82d3aa33bde33bc9b2f8bd728/oml/lightning/pipelines/train.py#L83

PR #363, opened 2023-04-28
Docs minor improvement 

AlekseySh:


AlekseySh:
I'll fix notebooks later. Seems like something weird happened on actions' side.

PR #381, opened 2023-06-03
Refactored ListDataset

AlekseySh:


AlekseySh:
`ListDataset` has been refactored to use `BaseDataset` inside

PR #382, opened 2023-06-04
Feature/metrics/stable metrics

dapladoc:
Introduce permutation stable version of the retrieval metrics. The idea is to sort `gt_tops` for which distances are the same, i.e. if `distances_tops` and `gt_tops` are

```python
distances_tops = [1, 2, 2, 2, 3]
gt_tops = [0, 1, 0, 1, 0]
```

then we need to obtain
```python
gt_tops = [0, 0, 1, 1, 0]
#             -------
#             sort gt_tops where there are multiple
#             distances of the same value
```

The code for the retrieval metrics itself doesn't change.

### DIFF
--- a/docs/source/oml/logging.rst
+++ b/docs/source/oml/logging.rst
@@ -17,7 +17,7 @@ There are two options to log your experiments when working with Pipelines:
 
   .. code-block:: bash
 
-      export NEPTUNE_API_TOKEN=your_token 
+      export NEPTUNE_API_TOKEN=your_token
       python train.py
 
 
--- a/oml/functional/metrics.py
+++ b/oml/functional/metrics.py
@@ -594,7 +594,7 @@ def _estimate_suitable_max_k(distances: Tensor, max_k: int) -> int:
         New value of ``max_k``
     """
     distances_tops, ii_top_k = torch.topk(distances, k=max_k + 1, largest=False)
-    n_same_distances = []
+    n_same_distances = [0]
     for i in torch.where(distances_tops[:, max_k - 1] == distances_tops[:, max_k])[0]:
         n_same_distances.append(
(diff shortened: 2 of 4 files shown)


dapladoc:
In the failed test called `test_metrics_is_similar_in_ddp` it is said that

> Metrics shouldn't be equal, but very similar

If I change the `atol` for the test from 5e-3 to 8e-3 the test passes. Is this workaround ok?

AlekseySh:
@dapladoc changing tolerance is okay

dapladoc:
@AlekseySh @DaloroAT hey guys. It looks like this PR is ready to be reviewed.

As I described above the main idea is to sort `gt_tops` values corresponding to the same distance values. In order to do it properly, first we need to find new value of `max_k` such that `torch.any(distances[:, max_k] == distances[:, max_k + 1] is False` (function `_estimate_suitable_max_k`). Next, we need to make the sorting (function `_sort_gt_tops`). And all the steps for getting the `gt_tops` is now encapsulated in function `_get_gt_tops`.

AlekseySh:
@DaloroAT What do you think? Does bringing the additional logic worth the benefits of metrics' stability? 
PS. Initially, I thought that the required extra logic would be smaller and simpler...

AlekseySh:
@dapladoc let's implement what we've decided offline:
1. measure overhead
2. introduce a flag for this functionality available for developers only (so, we don't expose is to config)
3. depends on the first step, set this flag as True or False by default

AlekseySh:
@dapladoc hey! just a reminder :)

AlekseySh:
Let me close it for now due to inactivity. Metrics code has also been significantly modified. So, anyway, the PR has to be reworked.

Issue #422, opened 2023-07-15
criterion.weight in CheckPoint resnet50


kodinkod:
When i want load my CheckPoint from train resnet50 with Arface i get    assert all(k.startswith(prefix) for k in state_dict.keys());
Becouse 'criterion.weight' in state_dict.  Can we find a bug fix and execute an extraction request: for example:
 if 'criterion.weight" in state_dict:
     del state_dict['criterion.weight']

in 128 row resnet.py

kodinkod:
I have not done pull requests before , I will be glad to cooperate and help

AlekseySh:
Hey, @kodinkod! First of all, thank you for using OML.

Before discussing a solution, could you please explain how did you create & train your model (please, provide code snippet or config.yaml). And also how did you load a state of your model? After that it will be more clear what should we do.

kodinkod:
for train:
`dataframe_name: /
dataset_root: /
logs_root:  log/
logs_folder: ${now:%Y-%m-%d_%H-%M-%S}_${postfix} 
show_dataset_warnings: True

num_workers: 12
cache_size: 0

transforms_train:
  name: augs_hypvit_torch
  args:
    im_size: 224

transforms_val:
  name: norm_resize_hypvit_torch
  args:
    im_size: 224
    crop_size: 224


metric_args:
  metrics_to_exclude_from_visualization: [cmc,]
  cmc_top_k: [1, 3, 5]
  map_top_k: [1, 3, 5]
  pfc_variance: [0.5, 0.9, 0.99]
  return_only_main_category: True
  visualize_only_main_category: True

log_images: True
metric_for_checkpointing: OVERALL/cmc/1
extractor:
  name: resnet
  args:
    arch: resnet50
    weights: resnet50_imagenet1k_v1
    normalise_features: True
    remove_fc: False
    gem_p: 1



criterion:
  name: arcface
  args:
    smoothing_epsilon: 0.0
    m: 0.4
    s: 31
    in_features: 1000 # for resnet
    num_classes: 821

sampler:  # batch sampler, in this case
  name: balance
  args:
    n_labels: 13
    n_instances: 4

bs_train: 60
bs_val: 256
max_epochs: 10000
valid_period: 5

optimizer:
  name: adam
  args:
    lr: 1e-5

# There are several options for logging:
logger:
  name: tensorboard
  args:
    save_dir: lightning_logs/arcface/
    log_model_checkpoints: True`

in python:
[...]

PR #430, opened 2023-08-06
fix criterion.weight in CheckPoint resnet50 #422

kodinkod:
i use :
def on_save_checkpoint(self, trainer: pl.Trainer, pl_module: pl.LightningModule, checkpoint) - Called when saving a checkpoint to give you a chance to store anything else you might want to save. I think it will be convenient to adjust the checkpoint.  

I had difficulties in testing and I have found only such a testing option so far, I will be glad of new knowledge and recommendations


AlekseySh:
Hey, @kodinkod! Thank you for working on this.

The problem with your solution is that the saving logic is linked with the logic of computing metrics. For example, it means that you cannot save/load checkpoint correctly if you don't calculate metric or if you calculate it differently (using another class).

AlekseySh:
As for the tests, here is a good example of how we test pipelines end-to-end: [link](https://github.com/OML-Team/open-metric-learning/tree/main/tests/test_runs/test_pipelines). But don't work on the test before we find a good solution.

AlekseySh:
I think that in general, it's okay to save the criterion's state, because it may be used later to restore the training. So, probably it's better to change the problematic function.

What if we change this
` assert all(k.startswith(prefix) for k in state_dict.keys())`
to something like this:
`assert all(k.startswith(prefix) for k in state_dict.keys() if not k.startswith('criterion'))`
?

It's a bit hacky, but better logically. 

I'm also open to other suggestions.

kodinkod:
I also thought that this would solve my problem, but with such a hacker solution I get 
```
self.model.load_state_dict(state_dict, strict=True)     
Error(s) in loading state_dict for ResNet:
        Unexpected key(s) in state_dict: "criterion.weight". 
```

`on_save_checkpoint`  just controls what is saved in State_dict;  

1. We can redefine `on_save_checkpoint`  in each metric class. 
2. We can somehow competently load each model by saving criterion.weight, and deleting it in state_dict so that we can use criterion.weight;

Also, I don't quite understand what  saved in `criterion.weight`

kodinkod:
If it is convenient for you, you can also write to me in telegam to discuss the problem: @kodlovecats

AlekseySh:
I think you got that error because `strict=True`, try to make it `False`.


> We can redefine on_save_checkpoint in each metric class.

Unfortunately, we cannot, because the idea is that users can have their own metrics. And, as I said before, probably I don't want to calculate any metric, but still want to save a checkpoint.

> We can somehow competently load each model by saving criterion.weight, and deleting it in state_dict so that we can use criterion.weight;

By the way, is that possible to control saving behavior on `pl.LightningModule` level, [here](https://github.com/OML-Team/open-metric-learning/blob/main/oml/lightning/modules/extractor.py#L14)?

> Also, I don't quite understand what saved in criterion.weight

Check [it](https://github.com/OML-Team/open-metric-learning/blob/a1a5d3e74baf15896b573ce47fd27d1a4b9924b4/oml/losses/arcface.py#L57) out.

kodinkod:
> I think you got that error because strict=True, try to make it False. 
yes, this fixes the problem;

AlekseySh:
I didn't suggest changing `strict` parameter from `True` to `False`, because I'm also worried that it may cause problems. I just wanted to double-check that we really understand what is going on. 

I think the optimal solution is to remove `criterion` key from the dict in all the existing models when we load them from the checkpoints. What do you think?

kodinkod:
It seems that only this method remains;  I think we can implement a function that will remove `criterion ` from the  `state_dict` and call it when loading the model. What do you think ?

AlekseySh:
@kodinkod Yes, I agree with you :) 

Check out this comment one more time before writing tests: https://github.com/OML-Team/open-metric-learning/pull/430#issuecomment-1670101780

AlekseySh:
@kodinkod Hey, are u going to finish this?

AlekseySh:
Just a friendly reminder

kodinkod:
sorry, I was busy with work, I'll try to do everything soon

AlekseySh:
@kodinkod no worries, ping we when the code is ready for review

kodinkod:
is the last commit a good practice? I would like to know this before I start porting similar code to other extractors

tests/test_oml/test_checpoints/test_checpoints_dict.py
```diff
@@ -0,0 +1,55 @@
+from pathlib import Path
```
AlekseySh, line 1: I think we don't need it anymore

AlekseySh:
@kodinkod I recommend you to run tests locally in a new env or in docker, so it will be easier to pass CI (check Makefile)

tests/test_oml/test_checpoints/test_checpoints_dict.py
kodinkod: yes

kodinkod:
@AlekseySh   If we're not in a rush I can continue working on it. It remains to fix a few extractors.

AlekseySh:
@kodinkod not in rush for sure, take your time

AlekseySh:
@kodinkod Hey! let's finish this! it shouldn't take too much time :)

kodinkod:
@AlekseySh I'm very sorry, I'm interested in participating in the project. but I've been incredibly busy at university and at work. I'm almost finished working on pr. I'm having difficulty writing tests for vit_clip and clip_unicom. I will be grateful for your help. Now I have little time and I'm afraid to make a mistake in a hurry.

thank you for the opportunity pr. I apologize again :(

AlekseySh:
@kodinkod thank you anyway! text me in telegram for cooperation (@lexibender)

PR #435, opened 2023-09-01
Two loaders example

leoromanovich:
add example for 2 validation loaders case
add case to tests
bug: found possible bug in train_val_pl.md, tests fail while log_images=True

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,6 @@ build_readme:
 	cat docs/readme/examples_source/extractor/train.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/val.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/train_val_pl.md >> ${README_FILE}
-	cat docs/readme/examples_source/extractor/train_2loaders_val.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/retrieval_usage.md >> ${README_FILE}
 	cat docs/readme/examples.md >> ${README_FILE}
 	# Pipelines
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ build_readme:
 	cat docs/readme/examples_source/extractor/train.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/val.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/train_val_pl.md >> ${README_FILE}
+	cat docs/readme/examples_source/extractor/train_2loaders_val.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/retrieval_usage.md >> ${README_FILE}
 	cat docs/readme/examples.md >> ${README_FILE}
 	# Pipelines
--- a/docs/readme/examples_source/extractor/train_2loaders_val.md
+++ b/docs/readme/examples_source/extractor/train_2loaders_val.md
@@ -52,3 +52,5 @@ print(metric_callback_2.metric.metrics)
 [comment]:lightning-2loaders-end
 </p>
 </details>
+
+<p></p>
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -30,6 +30,7 @@ build_readme:
 	cat docs/readme/examples_source/extractor/train.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/val.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/train_val_pl.md >> ${README_FILE}
+	cat docs/readme/examples_source/extractor/train_2loaders_val.md >> ${README_FILE}
 	cat docs/readme/examples_source/extractor/retrieval_usage.md >> ${README_FILE}
```
AlekseySh, line 34: we don't add all the examples to readme, and this particular example is too specific for the main readme

tests/test_runs/test_code_from_markdown.py
```diff
@@ -27,6 +27,11 @@ def find_code_block(file: Path, start_indicator: str, end_indicator: str) -> str
         ("extractor/val.md", "[comment]:vanilla-validation-start\n", "[comment]:vanilla-validation-end\n"),
         ("extractor/train_val_pl.md", "[comment]:lightning-start\n", "[comment]:lightning-end\n"),
         ("extractor/train_val_pl_ddp.md", "[comment]:lightning-ddp-start\n", "[comment]:lightning-ddp-end\n"),
+        (
```
AlekseySh, line 30: can you make it one line or it's black? if so, you can turn it off for this particular file

docs/readme/examples_source/extractor/train_2loaders_val.md
```diff
@@ -0,0 +1,59 @@
...
+
+optimizer = torch.optim.SGD(extractor.parameters(), lr=1e-6)
+criterion = TripletLossWithMiner(margin=0.1, miner=AllTripletsMiner())
+
+
+# 1st validation dataset
+val_dataset_1 = DatasetQueryGallery(df_val, dataset_root=dataset_root,
+                                  transform=get_normalisation_resize_torch(im_size=224))
+val_loader = torch.utils.data.DataLoader(val_dataset_1, batch_size=4)
+metric_callback = MetricValCallback(metric=EmbeddingMetrics(extra_keys=[val_dataset_1.paths_key, ],
```
AlekseySh, line 33: you can remove empty arguments for the metric (`[]`), I just did it for debugging

docs/readme/examples_source/extractor/train_2loaders_val.md
```diff
@@ -0,0 +1,59 @@
...
+
+print(metrics_lightning)
+print(metric_callback.metric.metrics)
+print(metric_callback_2.metric.metrics)
+```
+[comment]:lightning-2loaders-end
+</p>
+</details>
+
+[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1bVUgdBGWvQgCkba2YtaIRVlUQUz7Q60Z?usp=share_link)
```
AlekseySh, line 58: I was thinking about stopping creating collab for every example because they become outdated really fast and there is no easy way to test them. So, let's add this example without Colab

AlekseySh:
> bug: found possible bug in train_val_pl.md, tests fail while log_images=True

I think I remember something like that. The bug was reproducible only locally but CI/CD was okay (otherwise the main branch would be red). If you turn logging on again, I can run the tests and we will see what is going on.

Anyway, I will appreciate if you dig into that & understand what is going on even locally

tests/test_runs/test_code_from_markdown.py
leoromanovich: Yeah, it's black. Okay, i'll add ignore for this line.

docs/readme/examples_source/extractor/train_2loaders_val.md
leoromanovich: That's placeholder colab from previous example =) So yeah, I'll drop it out.

PR #448, opened 2023-10-15
Migration to torch and lightning 2.0

AlekseySh:
- [x] push new docker image (cpu, gpu)
- [x] fix `path_device` / `path_float`
- [x] check that DDP is actually improving speed
- [x] check models from ZOO can achieve expected metrics on benchmarks
- [x] double check `find_unused_parameters` behaviour 
- [x] green `all tests`
- [x] check `find_unused_parameters=True`

```
 Warning: find_unused_parameters=True was specified in DDP constructor, but did not find any unused parameters in the forward pass. This flag results in an extra traversal of the autograd graph every iteration,  which can adversely affect performance. If your model indeed never has any unused parameters in the forward pass, consider turning this flag off. Note that this warning may be a false positive if your model has flow control causing later iterations to have unused parameters. (function operator())
```


AlekseySh:
Note, I've patched CI file so it's working for `all tests` ^
I will roll it back later

ci/requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy>=1.17.2, <1.23.1
-torch>=1.10, <=1.13.0
-torchvision>=0.11, <=0.14.0
-pytorch-lightning>=1.5.9, <1.7.0  # refactor code, if you want newer versions
+torch>=2.0
+pytorch-lightning>=2.1.0
```
DaloroAT, line 3: pytorch-lightning follow torch version, so pytorch-lightning>=2.0

oml/lightning/callbacks/metric.py
```diff
@@ -54,11 +55,14 @@ def __init__(
         self._collected_samples = 0
         self._ready_to_accumulate = False
 
-    def _calc_expected_samples(self, trainer: pl.Trainer, dataloader_idx: int) -> int:
-        return self.samples_in_getitem * len(trainer.val_dataloaders[dataloader_idx].dataset)
+    def _calc_expected_samples(self, trainer: pl.Trainer, dataloader_idx: int = 0) -> int:
```
DaloroAT, line 58: any reason to default values for dataloader_idx? The same for other functions

DaloroAT (approved):

oml/lightning/callbacks/metric.py
AlekseySh: it fails otherwise, so they must be here

ci/requirements.txt
AlekseySh: okay, let's try to run tests for lightning >= 2.0

AlekseySh: in CI

AlekseySh: seems like it works

PR #454, opened 2023-11-22
Add groups of samples so samples from the same group are ignored during the evaluation

AlekseySh:
This feature is useful for video-based re-id tasks, where we may get too optimistic metrics because the query frame has too many similar frames in the gallery that will be returned at the top of the retrieval results. To avoid this, we introduce groups so samples from the same group are ignored during the retrieval-style evaluation.
In other words, we ignore trivial matches during the evaluation.

ToDo:
- [x] tests
- [x] add to config API for train / val
- [x] update dataset

For future:
_Improve training: get rid of trivial positive pairs in constrastive losses._


AlekseySh:
cc @leoromanovich

PR #470, opened 2023-12-06
Optional GradCam

AlekseySh:


oml/models/resnet/extractor.py
```diff
@@ -174,6 +172,10 @@ def draw_gradcam(self, image: Union[np.ndarray, TPILImage]) -> Union[np.ndarray,
         .. _GradCam: https://arxiv.org/abs/1610.02391
 
         """
+        # this is the optional dependency
```
DaloroAT, line 175: I prefer to move this comment from code to traceback :-) Example of how to modify traceback message: 
https://github.com/DaloroAT/first_breaks_picking/blob/5804bfc112d16d1aa38c0acf975daa992960cae7/first_breaks/data_models/independent.py#L160

But it's up to you

DaloroAT (approved):
I prefer to add more information about dependencies

PR #473, opened 2023-12-06
Rework tests

AlekseySh:


AlekseySh:
**Changelog**: 
* Starting now we will be running tests for all the supported pythons, instead of only for 3.8 to have better coverage. We will run short_tests for `pull_request` and long tests for `push` requests. Tests are running outside of docker, on the CI Ubuntu machine.
* To avoid excessive checkpoints downloading from Zoo for every Python version, this test is executed only when a special global env is provided: `DOWNLOAD_ZOO_IN_TESTS=yes`. So, we provide this env only for one of the Python versions from the whole list.
* We keep all_tests running in docker only for one Python version and only for `push` requests. It's done to avoid the need to maintain ~3-5 docker containers for every Python version and to save the time needed for every docker build.

* Marked a number of tests as long to make short_tests even shorter (not all of them are really "long", but it seems unlikely that we need to run all of them on every commit)
* Requirements needed for testing (pre-commit and pytest) are in one place and re-used in workflows & dockerfile

.github/workflows/docker_tests.yaml
```diff
@@ -0,0 +1,23 @@
+name: Tests in Docker (3.8)
+
+on:
+  pull_request:
```
DaloroAT, line 4: You can rm it since you have it only for `push` event. Or do you want to see a list of all project tests?

AlekseySh: yeah, I want to see the whole list

PR #480, opened 2023-12-08
Simplified working with benchmark datasets

AlekseySh:


leoromanovich:
I'll check if all works out of the box on my machine.

leoromanovich:
- [x] CARS196 with/without bboxes - all good
- [x] CUB_200_2011 with/without bboxes - all good
- [x] In_Shop with/without bboxes - all good
- [x] SOP without bboxes - all good

- I think it's better to replace wandb loggers in all examples to tb (in case if user don't have an account there).
- It looks like in_shop needs a V100 or higher to run (I believe in case of reproducibility). What do you think about reducing hardware requirements to be able to run it locally on consumer gpu (like 4090 or lower)? Same for SOP example.

AlekseySh:
@leoromanovich thx for checking! I'm merging the PR then

As for loggers - I would like to keep examples that show that not only wandb logger can be user
As for hardware -- you are right, the parameters there allow us to reproduce the examples from Zoo. Anyway, we cannot gues what is users hardware, so they need to adjust the batch size on their own

PR #493, opened 2023-12-31
Update README.md

eltociear:
separe -> separate


AlekseySh:
Hey, @eltociear, thank you for your interest in OML!

We have an auto-generated readme, so the changes have to be made in another place.
From our contribution [guide](https://open-metric-learning.readthedocs.io/en/latest/oml/contributing.html):

```
If you want to change README.md you should go to docs/readme, change the desired section and then build readme via make build_readme. So, don’t change the main readme file directly, otherwise tests will fail.
```

Issue #509, opened 2024-03-20
Add MLFlowLogger


deepslug:
I've just started using OML and I absolutely love it! (More questions are on the way...)

It'd be nice if it supported MLFlowLogger.
https://lightning.ai/docs/pytorch/stable/extensions/generated/lightning.pytorch.loggers.MLFlowLogger.html

Fissium:
@deepslug Use can use the following code example to add the mlflow logger to your pipeline:

```python
import logging
import warnings

import hydra
from omegaconf import DictConfig
from oml.const import HYDRA_VERSION
from oml.lightning.pipelines.train import extractor_training_pipeline
from oml.registry.loggers import LOGGERS_REGISTRY
from pytorch_lightning.loggers import MLFlowLogger

logging.basicConfig(level=logging.INFO)

logger = logging.getLogger(__name__)

warnings.filterwarnings("ignore")


LOGGERS_REGISTRY["mlflow"] = MLFlowLogger


@hydra.main(
    config_path="./configs/",
    config_name="train_vit_mlflow.yaml",
    version_base=HYDRA_VERSION,
)
def main_hydra(cfg: DictConfig) -> None:
    logger.info("Training model.")
    extractor_training_pipeline(cfg)


if __name__ == "__main__":
    main_hydra()
```

Then add something similar to your config:

```yaml
logger:
  name: mlflow
  args:
    experiment_name: "ViT_oml"
    run_name: "vit_log_artifacts"
```

AlekseySh:
@deepslug Thank you for using OML :) Waiting for your further questions

@Fissium Thank you so much for such a great answer! It's a valid approach, but it requires changing source code of the library (as you mentioned in the "final step"). To avoid this and make it available from pip, we need someone who can add this functionality to the repo. So, contributions are welcome :)

deepslug:
@Fissium 
Thank you for your detailed answer! I've actually tried something similar and realized I had to modify parser.py, hence my question.

@AlekseySh 
I'll post my question when I have everything sorted out.

AlekseySh:
@deepslug , great, thank you

AlekseySh:
@armored-guitar is going to work on this issue.

PR #516, opened 2024-03-27
Refactored loggers

AlekseySh:
* unified logging interfaces
* get rid of `if isinstance(...)` in `MetricsCallback`
* added cloud logging to tests (only to "all tests" on python 3.8)
* get rid of logging code in config parser 
* updated neptune version


AlekseySh:
first step before https://github.com/OML-Team/open-metric-learning/issues/509

PR #518, opened 2024-03-28
use torchvision.transforms.v2

Anton-Cherepkov:
self-explanatory


AlekseySh:
Hey, @Anton-Cherepkov 
Thank you for your contribution. I run CI tests.

Though, I'm not sure if we should replace v1 transforms with v2 instead of keeping both of them.

AlekseySh:
@Anton-Cherepkov , hey! If you are still interested in the contribution, please, make it up to date with the current main and let's move on with this PR.

PR #519, opened 2024-03-29
Add DINOv2 model

Fissium:
DINOv2 implementation

By default, it uses F.scaled_dot_product_attention, if possible.
It can be disabled by setting the environment variable `USE_FUSED_ATTN=0`.

It also looks like we can add the model to the existing ViTExtractor pipeline.

**TODO**: Evaluate the model on benchmarks and it to Zoo table. **DONE**


oml/models/vit_dinov2/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+)
+from oml.utils.io import download_checkpoint_one_of
+from oml.utils.misc_torch import normalise
+
+_DINOV2_BASE_URL = "https://dl.fbaipublicfiles.com/dinov2"
+
+
+class ViTExtractor_v2(IExtractor):
+    """
+    The base class for the extractors that follow VisualTransformer architecture.
```
AlekseySh, line 27: Seems like you forgot to remove it

oml/models/vit_dinov2/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+            x = normalise(x)
+
+        return x
+
+    @property
+    def feat_dim(self) -> int:
+        return len(self.model.norm.bias)
+
+
+__all__ = ["ViTExtractor_v2"]
```
AlekseySh, line 123: All the code above looks like the same code as we already have in `ViTExtractor`.
So, instead of having `ViTExtractor_v2` we may have only `ViTExtractor` with extended `pretrained_models`.
Correct me if I'm wrong.

If so, you may place the needed external code under `vit_dino/external_v2`

What do you think?

AlekseySh (changes requested):
Thank you for the contribution! Let's discuss details

oml/models/vit_dinov2/extractor.py
Fissium: Yes, definitely

oml/models/vit_dinov2/extractor.py
Fissium: Much better than the separate class "ViTExtractor_v2". I'll fix it.

docs/readme/zoo.md
```diff
@@ -29,6 +29,12 @@ The metrics below are for **224 x 224** images:
...
 |    `ResnetExtractor.from_pretrained("resnet50_moco_v2")`     |          0.493           |       0.267        |    0.264     |  0.149   |
 | `ResnetExtractor.from_pretrained("resnet50_imagenet1k_v1")`  |          0.515           |       0.284        |    0.455     |  0.247   |
+|      `ViTExtractor_v2.from_pretrained("vits14_dinov2")`      |          0.566           |       0.334        |    0.797     |  0.503   |
+|   `ViTExtractor_v2.from_pretrained("vits14_reg_dinov2")`     |          0.566           |       0.332        |    0.795     |  0.740   |
+|      `ViTExtractor_v2.from_pretrained("vitb14_dinov2")`      |          0.565           |       0.342        |    0.842     |  0.644   |
+|   `ViTExtractor_v2.from_pretrained("vitb14_reg_dinov2")`     |          0.557           |       0.324        |    0.833     |  0.828   |
+|      `ViTExtractor_v2.from_pretrained("vitl14_dinov2")`      |          0.576           |       0.352        |    0.844     |  0.692   |
+|   `ViTExtractor_v2.from_pretrained("vitl14_reg_dinov2")`     |          0.571           |       0.340        |    0.840     |  0.871   |
```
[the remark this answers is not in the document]
Fissium: See bb13cc04f2e108e2593c6b7d0fd5dddb283357ad

oml/models/vit_dinov2/extractor.py
Fissium: See fcd5cdd6ed4d5d9f2ca4ac9b39afcb4534a357c1

oml/models/vit_dinov2/extractor.py
Fissium: No need to fix because of fcd5cdd6ed4d5d9f2ca4ac9b39afcb4534a357c1

AlekseySh:
@Fissium Great job, let's run CI

AlekseySh (dismissed):
LGTM!

PR #540, opened 2024-04-17
Removing duplicates and preserving order in Accumulator (DDP)

AlekseySh:


Fissium (commented):
Hi. I ran several pipelines in `DDP` mode from the `accumulator_duplicates` branch to check potential differences in metrics compared to the `main` branch. Fortunately, the metrics remain identical across both branches.

oml/metrics/embeddings.py
```diff
@@ -161,8 +161,17 @@ def setup(self, num_samples: int) -> None:  # type: ignore
 
         self.acc.refresh(num_samples=num_samples)
 
-    def update_data(self, data_dict: Dict[str, Any]) -> None:  # type: ignore
-        self.acc.update_data(data_dict=data_dict)
+    def update_data(self, data_dict: Dict[str, Any], indices: Optional[List[int]] = None) -> None:  # type: ignore
+        """
+        Args:
+            data_dict: Batch of data containing records of the same size: ``bs``.
+            indices: Global indices of the elements in your records withing the range of ``(0, dataset_size - 1)``.
```
DaloroAT, line 168: typo: withing

oml/metrics/embeddings.py
```diff
@@ -382,5 +391,9 @@ class EmbeddingMetricsDDP(EmbeddingMetrics, IMetricDDP):
     def sync(self) -> None:
         self.acc = self.acc.sync()
 
+    def update_data(self, data_dict: Dict[str, Any], indices: List[int]) -> None:  # type: ignore
+        # indices are obligatory in DDP, so we don't accumulate shuffled data
```
DaloroAT, line 395: Insufficient comment. The problem is not in shuffling, but in duplicates (at least shuffling is ok in a single device for some reason)

oml/metrics/embeddings.py
```diff
@@ -161,8 +161,17 @@ def setup(self, num_samples: int) -> None:  # type: ignore
 
         self.acc.refresh(num_samples=num_samples)
 
-    def update_data(self, data_dict: Dict[str, Any]) -> None:  # type: ignore
-        self.acc.update_data(data_dict=data_dict)
+    def update_data(self, data_dict: Dict[str, Any], indices: Optional[List[int]] = None) -> None:  # type: ignore
```
DaloroAT, line 164: Why are indices not obligatory? Do you want to do it in the future with better integration with datasets where order-keeping is important?

oml/utils/misc_torch.py
```diff
@@ -126,31 +124,33 @@ def _check_is_sequence(val: Any) -> bool:
         return False
 
 
-def drop_duplicates_by_ids(ids: List[Hashable], data: Tensor, sort: bool = True) -> Tuple[List[Hashable], Tensor]:
+TData = Tuple[List[Any], BoolTensor, FloatTensor, LongTensor, Tensor, np.ndarray]
```
DaloroAT, line 127: Redundant types?
~BoolTensor, FloatTensor, LongTensor~, Tensor
or
BoolTensor, FloatTensor, LongTensor, ~Tensor~

tests/test_oml/test_ddp/test_accumulator.py
```diff
@@ -12,31 +12,61 @@
...
+@pytest.mark.parametrize("create_duplicate", [True, False])
+def test_ddp_accumulator(world_size: int, device: str, create_duplicate: bool) -> None:
+    run_in_ddp(world_size=world_size, fn=check_ddp_accumulator, args=(device, create_duplicate))
 
 
-def check_ddp_accumulator(rank: int, world_size: int, device: str) -> None:
+@pytest.mark.parametrize("device", ["cpu", "cuda"] if torch.cuda.is_available() else ["cpu"])
+@pytest.mark.parametrize("create_duplicate", [True, False])
+def test_fake_ddp_accumulator(device: str, create_duplicate: bool) -> None:
+    # we expect the same behaviour without initializing DDP
```
DaloroAT, line 23: What is the goal of this setup?

Maybe you expect `the same behaviour without initializing DDP`, but I think [these guys](https://github.com/OML-Team/open-metric-learning/blob/ec091f183fc1f7182a4c28e9073bc1a01b7fa8cc/oml/ddp/utils.py#L7) from `torch` that `Accumulator` use expects a bit different setup when working with DDP 😁

oml/metrics/accumulation.py
```diff
@@ -1,16 +1,17 @@
...
 import torch
-from torch.distributed import get_world_size
+from torch import BoolTensor, FloatTensor, LongTensor, Tensor
 
-from oml.ddp.utils import is_ddp, sync_dicts_ddp
+from oml.ddp.utils import get_world_size_safe, sync_dicts_ddp
+from oml.utils.misc_torch import unique_by_ids
 
-TStorage = Dict[str, Union[torch.Tensor, np.ndarray, List[Any]]]
+TStorage = Dict[str, Union[FloatTensor, BoolTensor, LongTensor, Tensor, np.ndarray, List[Any]]]
```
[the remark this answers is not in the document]
AlekseySh: Agree, just tensor is okay and shorter

oml/metrics/accumulation.py
```diff
@@ -1,16 +1,17 @@
...
 
 
 class Accumulator:
-    def __init__(self, keys_to_accumulate: Sequence[str]):
+    def __init__(self, keys_to_accumulate: Tuple[str, ...]):
```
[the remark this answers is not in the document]
AlekseySh: For safety. I thinks it's better to pass immutable argument and store it in the field. Also, typeching works better with more narrow types.

oml/metrics/accumulation.py
```diff
@@ -75,20 +80,36 @@ def _put_in_storage(self, key: str, batch_value: Any) -> None:
...
+                     will remove accumulated duplicates and return the elements in the sorted order after ``.sync()``.
+                     Indices may be useful in DDP (because data is gathered shuffled, additionally you may also get
+                     some duplicates due to padding). In the single device regime it's also useful if you accumulate
+                     data in shuffled order.
 
         """
-        bs_values = [len(data_dict[k]) for k in self.keys_to_accumulate]
+        keys = list(self.keys_to_accumulate)
+
+        if indices is not None:
```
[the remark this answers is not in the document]
AlekseySh: agree

oml/metrics/accumulation.py
```diff
@@ -27,12 +28,16 @@ def __init__(self, keys_to_accumulate: Sequence[str]):
...
+        self.indices_key = "element_indices"  # internal key to keep track of elements order if provided
+
     def refresh(self, num_samples: int) -> None:
         """
         This method refreshes the state.
 
         Args:
-            num_samples:  The total number of elements you are going to collect (for memory allocation)
+            num_samples:  The total number of elements you are going to collect (for memory allocation).
+                          If you've provided ``ids`` in ``self.update_data`` method, this number may decrease
```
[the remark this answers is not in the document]
AlekseySh: okay, removed

oml/metrics/accumulation.py
```diff
@@ -27,12 +28,16 @@ def __init__(self, keys_to_accumulate: Sequence[str]):
         self._collected_samples = 0
         self._storage: TStorage = dict()
 
+        self.indices_key = "element_indices"  # internal key to keep track of elements order if provided
```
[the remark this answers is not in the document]
AlekseySh: done (the rest private fields have 1 underscores, so I've also added one)

oml/metrics/embeddings.py
AlekseySh: fixed

oml/metrics/embeddings.py
AlekseySh: yes, that's right
I've added `todo 522` comment to make sure I will not forget

oml/utils/misc_torch.py
AlekseySh: Agree, prefer the first option

tests/test_oml/test_ddp/test_accumulator.py
AlekseySh: The goal of this setup is to check that duplicates can be removed in non DDP setup as well. For example, internal implementation of Accumulator may immediately remove `self` in `sync` method if `not is_ddp()` as it was before. However as far as we want to remove duplicated in any regime, we need to have a test for it.

AlekseySh: anyway, changed comment a bit:
` # we expect the same duplicate removing behaviour without initializing DDP`

oml/metrics/embeddings.py
AlekseySh: okay, changed the comment

oml/metrics/accumulation.py
```diff
@@ -75,20 +80,36 @@ def _put_in_storage(self, key: str, batch_value: Any) -> None:
...
+                     Indices may be useful in DDP (because data is gathered shuffled, additionally you may also get
+                     some duplicates due to padding). In the single device regime it's also useful if you accumulate
+                     data in shuffled order.
 
         """
-        bs_values = [len(data_dict[k]) for k in self.keys_to_accumulate]
+        keys = list(self.keys_to_accumulate)
+
+        if indices is not None:
+            assert isinstance(indices, List)
```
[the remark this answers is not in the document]
AlekseySh: Fortunately, this bug is caught by another assertion: we compare length of collected data and lengths of ids, but it happens only during the sync. So, I agree that we should raise an error earlier. 

I converted your example to a test case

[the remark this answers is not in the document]
DaloroAT: > collected data and lengths of ids
Not exactly. We use this method for assert:
```python
    def is_storage_full(self) -> bool:
        return self.num_samples == self.collected_samples
```
where `self.collected_samples` is incremented by `len` of each batch, not `indices`.

Anyway, with extra assertion it will work as expected.

tests/test_oml/test_ddp/test_accumulator.py
DaloroAT: Got it. 
But I still have concern about *partially* initialized DDP. `check_ddp_accumulator` call `init_ddp` on first line where DDP process group is created. So in this case maybe `is_ddp` is `True`, but test works because `world_size` is 1. Could you add print and check it?

oml/metrics/accumulation.py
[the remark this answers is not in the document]
DaloroAT: I told about value, like `"__element_indices"`. Because user will see it even if he didn't asked it to collect in `key_to_accumulate`.
Underscore for attribute is also ok for consistency

tests/test_oml/test_ddp/test_accumulator.py
AlekseySh: i dont understand you
we don't run init_ddp in this test, we only run it for the real ddp test

oml/metrics/accumulation.py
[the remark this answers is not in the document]
AlekseySh: oh, okay, did it

tests/test_oml/test_ddp/test_accumulator.py
AlekseySh: check where exactly ddp_init is

oml/metrics/accumulation.py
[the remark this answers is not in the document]
AlekseySh: hm okay

PR #547, opened 2024-04-26
Refactoring integration

AlekseySh:


tests/test_oml/test_metrics/test_embedding_metrics.py
```diff
@@ -126,33 +111,30 @@ def run_retrieval_metrics(case) -> None:  # type: ignore
...
+        calc.update(embeddings=batch[dataset.input_tensors_key], indices=batch[dataset.index_key])
 
     metrics = calc.compute_metrics()
 
     compare_dicts_recursively(gt_metrics, metrics)
 
     # the euclidean distance between any one-hots is always sqrt(2) or 0
-    assert compare_tensors_as_sets(calc.distance_matrix, torch.tensor([0, math.sqrt(2)]))
+    assert (
+        torch.isclose(calc.retrieval_results.distances, torch.tensor([0.0])).any()
```
AlekseySh, line 133: the logic has changed a bit because now we only have distances to the closest items, and don't have access to the whole distance matrix

thus, we cannot guarantee we meet all the potential distances values as before, but what we actually meet must much these potential values

tests/test_oml/test_metrics/test_embedding_visualizations.py
```diff
@@ -1,70 +0,0 @@
-from functools import partial
```
AlekseySh, line 1: this logic moved to RetrievalResults

oml/metrics/embeddings.py
```diff
@@ -5,85 +5,67 @@
...
 
 TMetricsDict_ByLabels = Dict[Union[str, int], TMetricsDict]
 
 
-def validate_dataset(mask_gt: Tensor, mask_to_ignore: Tensor) -> None:
-    is_valid = (mask_gt & ~mask_to_ignore).any(1).all()
-
-    if not is_valid:
-        raise RuntimeError("There are queries without available correct answers in the gallery!")
+def calc_retrieval_metrics_rr(
```
AlekseySh, line 35: I put it here to avoid circular imports

docs/readme/examples_source/postprocessing/predict.md
```diff
@@ -4,15 +4,14 @@
...
-import torch
 
 from oml.datasets import ImageQueryGalleryDataset
 from oml.inference import inference
 from oml.models import ConcatSiamese, ViTExtractor
 from oml.registry.transforms import get_transforms_for_pretrained
 from oml.retrieval.postprocessors.pairwise import PairwiseReranker
 from oml.utils.download_mock_dataset import download_mock_dataset
-from oml.utils.misc_torch import pairwise_dist
+from oml.retrieval import RetrievalResults
```
DaloroAT, line 14: [nit] looks like this import should be a few lines above after sorting. Maybe not only in this file

oml/metrics/embeddings.py
```diff
@@ -148,114 +115,79 @@ def __init__(
...
+        self._acc_embeddings_key = "__embeddings"
+        self.acc = Accumulator(keys_to_accumulate=(self._acc_embeddings_key,))
+
+    def setup(self, num_samples: Optional[int] = None) -> None:  # type: ignore
+        self.retrieval_results = None
         self.metrics = None
+        self.metrics_unreduced = None
 
-        self.acc.refresh(num_samples=num_samples)
+        num_samples = num_samples if num_samples is not None else len(self.dataset)
```
DaloroAT, line 126: Difficult logic 🤔 

On the one hand, metric gets the dataset in constructor, so it always has to be `len(self.dataset)`. On the other hand, such logic is invalid for DDP setup...

oml/metrics/embeddings.py
```diff
@@ -148,114 +115,79 @@ def __init__(
...
-            ii_aligned = list(range(len(self.dataset)))
-            assert ii_aligned == self.acc.storage[INDEX_KEY].tolist(), "The data is shuffled!"  # type: ignore
-            self.distance_matrix = self.postprocessor.process(self.distance_matrix, dataset=self.dataset)
+            self.retrieval_results = self.postprocessor.process(self.retrieval_results, self.dataset)
 
     def compute_metrics(self) -> TMetricsDict_ByLabels:  # type: ignore
+        self.acc = self.acc.sync()  # if DDP gathering happens here
+
+        # todo 522: put back fnmr metric
+
```
DaloroAT, line 159: Error below is not necessary since the same check happens in `acc.sync()`

oml/metrics/embeddings.py
```diff
@@ -276,6 +208,9 @@ def compute_metrics(self) -> TMetricsDict_ByLabels:  # type: ignore
 
         return metric_to_return  # type: ignore
 
+    def ready_to_visualize(self) -> bool:
+        return isinstance(self.dataset, IVisualizableDataset)
```
DaloroAT, line 212: suppose metrics also should be calculated. At least `def visualize(self)` uses metrics

oml/metrics/embeddings.py
DaloroAT: I have a mixed feeling... If it's optional, it looks redundant to call this function without args in the code. With optional it's more natural to call it implicitly during the first call of `update` or explicitly in the DDP code.

DaloroAT:
>Removed IMetricDDP, EmbeddingMetricsDDP. Reason: having EmbeddingMetrics is enough, because we do accumulator sync there anyway.

Looks like you also can remove `MeticValCallbackDDP` by moving a few lines to `MeticValCallback`

oml/metrics/embeddings.py
AlekseySh: not sure how I can implement this...

AlekseySh:
> Looks like you also can remove MeticValCallbackDDP by moving a few lines to MeticValCallback

I will add this to todo

PR #548, opened 2024-04-26
Update _read_bytes function in images.py support for URL

stoensin:
Add support for fetching images from URLs in _read_bytes_image

Previously, the function only supported reading image files from local paths. This commit extends its functionality to also fetch images from URLs if the path starts with 'http://' or 'https://'. The function now uses the 'requests' library to download the image content when a URL is provided.

If the URL fetch is successful, the image content is returned as bytes. If the status code is not 200, a ValueError is raised with a message indicating the failure.

The existing file reading functionality remains unchanged for non-URL paths, ensuring backward compatibility.

Changes:
- Modified _read_bytes_image to check if the path is a URL
- Added conditional logic to handle URL paths using requests.get
- Included error handling for failed URL fetch attempts

For now, we don't push you to follow any predefined schema of issue, but ensure you've already read our contribution guide: https://open-metric-learning.readthedocs.io/en/latest/from_readme/contributing.html.

### DIFF
--- a/oml/datasets/images.py
+++ b/oml/datasets/images.py
@@ -157,8 +157,17 @@ class ImageBaseDataset(IBaseDataset, IVisualizableDataset):
 
     @staticmethod
     def _read_bytes(path: Union[Path, str]) -> bytes:
-        with open(str(path), "rb") as fin:
-            return fin.read()
+        if isinstance(path, str) and path.startswith(('http://', 'https://')):
+            # If the path is a URL, use requests to fetch the content
+            response = requests.get(path)
+            if response.status_code == 200:
+                return response.content
+            else:
+                raise ValueError(f"Failed to fetch image from URL: {path}")
+        else:
+            # Otherwise, assume it's a local file path
+            with open(str(path), "rb") as fin:
+                return fin.read()
 
     def __getitem__(self, item: int) -> Dict[str, Union[FloatTensor, int]]:
         img_bytes = self.read_bytes(self._paths[item])


AlekseySh:
Hey, @stoensin ! Thank you for the interest in OML.

I think we need some tests for this. As an input you can use this url: `https://i.ibb.co/wsmD5r4/photo-2022-06-06-17-40-52.jpg`. You can put it here: https://github.com/OML-Team/open-metric-learning/tree/main/tests/test_oml/test_datasets

oml/datasets/images.py
```diff
@@ -157,8 +157,17 @@ def __init__(
 
     @staticmethod
     def _read_bytes(path: Union[Path, str]) -> bytes:
-        with open(str(path), "rb") as fin:
-            return fin.read()
+        if isinstance(path, str) and path.startswith(('http://', 'https://')):
+            # If the path is a URL, use requests to fetch the content
+            response = requests.get(path)
```
DaloroAT, line 162: From a performance point of view, it would be more efficient to create a session instance `self.session = self.session or requests.Session()` and make requests using `self.session.get(...)`. Otherwise, for each call of `__getitem__`, you will need to establish a new connection with the remote server (DNS resolution, certificate exchange, etc). Establishing a connection can take a huge amount of time.

I'm not sure about an exact implementation but doing `requests.get` is not a good idea for dataset scenario, because we can reuse a session.

AlekseySh:
Hey, @stoensin ! Let's move on with this PR. Please, see the comment above: https://github.com/OML-Team/open-metric-learning/pull/548#discussion_r1596831798

AlekseySh:
@stoensin please, re-open this PR, when you ready to continue

Issue #535, opened 2024-04-14
Make shortcuts for imports


AlekseySh:
`from oml.datasets.base import X` -> `from oml.datasets import X`
So, make all the functions and classes accessible with no more than two-level imports

mimi030:
Hello! I'm interested in contributing to this issue. Is it still open?

Could you confirm if the goal here is to simplify imports and enhance readability by reducing import depth across all affected files?

AlekseySh:
Hey, @mimi030 
Thank you for the interest!

That's correct. Ideally, I want to have no more than 2 levels in our imports. For example,
`from oml.losses.triplet import TripletLossWithMiner` -> `from oml.losses import TripletLossWithMiner`.
Also, don't forget to update examples (check the contribution guide before, please).

I think, you can start with losses, models and miners. I would'n touch validation part much at the moment because we are expecting big updates there.

mimi030:
Thanks for clarifying the scope, @AlekseySh ! 

I'll start by simplifying import paths within the 'losses,' 'models,' and 'miners' modules to ensure they don't exceed two levels. For example, I'll refactor import statements like this:

`# Example from oml/models/vit_unicom
/extractor.py
# Before refactoring
from oml.interfaces.models import IExtractor
from oml.models.utils import (
    remove_criterion_in_state_dict,
    remove_prefix_from_state_dict,
)
from oml.models.vit_unicom.external import vision_transformer
from oml.models.vit_unicom.external.model import load  # type: ignore
from oml.utils.misc_torch import normalise

# After refactoring
from oml.interfaces import IExtractor
from oml.models import (
    remove_criterion_in_state_dict,
    remove_prefix_from_state_dict,
    vision_transformer,
    load,  # type: ignore
)
from oml.utils import normalise
`

Could you also clarify which components are included in the 'validation part' you mentioned? Is it related to the extractor_training_pipeline and extractor_validation_pipeline, or the validation steps outlined in the [documentation](https://open-metric-learning.readthedocs.io/en/latest/feature_extraction/pipelines.html#validation)? Thanks for your help!

I'll also update relevant examples in the documentation to reflect these changes.

AlekseySh:
Your example looks great!

By validation part I mean postprocessors, metrics and so on. So, that is why I suggest to start with losses, miners, samplers, and models. You can simplify their imports everywhere.

mimi030:
[...]

PR #554, opened 2024-04-30
Make imports shortcuts

mimi030:
**Amendments Overview:** 
This pull request introduces the following improvements:

- Simplified import statements in the 'losses' module.
- Simplified imports from the 'miners' module.
- Simplified import paths for the 'models' module.
- Simplified import statements in the 'samplers' module.
- Simplified import examples in the 'Features Extraction' documentation.
- Simplified imports from 'losses', 'models', 'miners', and 'samplers' within the '/oml' module.
- Added dependencies installation info to contribution guide

**Notes:** 
By moving import statements into the corresponding modules' init.py files, most imports are simplified. A few import statements remain unchanged to prevent circular import issues.

Additionally, the import examples in the 'Features Extraction' documentation in Colab require permission to proceed, so they remain unchanged.


AlekseySh:
Hey, @mimi030 
After you solve the conflicts, we can run CI/CD

AlekseySh:
Also, please, double check contribution pipeline & see how to update Readme

AlekseySh:
And just to make sure we are on the same page: it's not necessary to simplify the imports inside the modules. I mean, you can do it, but the final purpose if to simplify examples, so, you can only simplify imports there.

mimi030:
> And just to make sure we are on the same page: it's not necessary to simplify the imports inside the modules. I mean, you can do it, but the final purpose if to simplify examples, so, you can only simplify imports there.

So just to clarify, should I focus only on updating the import statements in the examples found under [Features Extraction](https://open-metric-learning.readthedocs.io/en/latest/feature_extraction/python_examples.html)? I'm feeling a bit confused about the task scope now, as I had understood from our discussion [here](https://github.com/OML-Team/open-metric-learning/issues/535#issuecomment-2065410584) that we were looking to implement these changes more broadly.

To ensure I'm on the right track, could you please specify which specific files I should be targeting for these updates?

Thank you for your guidance!

mimi030:
> Also, please, double check contribution pipeline & see how to update Readme

Certainly! I've previously gone through the Contributing in general and Contributing to documentation, but I'll be happy to double-check everything again to ensure everything is in order. If you have specific areas or updates you'd like me to focus on, please let me know.

AlekseySh:
from the guide:
```
## Contributing to documentation
* If you want to change `README.md` you should go to `docs/readme`, change the desired section and then build
  readme via `make build_readme`. *So, don't change the main readme file directly, otherwise tests will fail.*

```

PR #560, opened 2024-05-12
Remove MetricValCallbackDDP and samples_in_getitem

AlekseySh:
* `MetricValCallback` is enough to handle DDP
* `samples_in_getitem` is not used


AlekseySh:
I think, there is nothing to review, but just for your information, @DaloroAT

PR #563, opened 2024-05-12
RetrievalResults as tensors with padding

AlekseySh:


AlekseySh:
I've decided to use Sequence of Tensors for RetrievalResults: https://github.com/OML-Team/open-metric-learning/pull/565/files

1. The code is easier to maintain 
2. Metric calculations and other manipulations are fast enough

PR #565, opened 2024-05-18
RetrievalResults as sequence of tensors

AlekseySh:
**CHANGELOG**

* `RetrievalResults` uses Sequence of Tensors which may have different size. In other words, it allows us to support the case when queries have different number of retrieved items.
* Consequently, changed `batched_knn`, `retrieval_metrics` and `PairwiseReranker` to support new input type.
* Added assert that distances arrive sorted to `RetrievalResults`, retrieved ids are unique and other checks. 

New tests:
* Added tests on corner cases for `RetrievalResults` creation.
* Added tests on visualization when queries in `RetrievalResults` have different number of retrieved items.
* Added new test with predefined values for `batched_knn` to make debugging easier.
* Changed existing postprocessor tests: used `sequence` in datasets so queries have different number of retrieved items and we actually test new functionality.

@leoromanovich and I also checked that using Sequence of Tensors doesn't lead to poor performance on validation.


oml/retrieval/retrieval_results.py
```diff
@@ -24,34 +26,46 @@
 class RetrievalResults:
     def __init__(
         self,
-        distances: FloatTensor,
-        retrieved_ids: LongTensor,
-        gt_ids: List[LongTensor] = None,
+        distances: Sequence[FloatTensor],
+        retrieved_ids: Sequence[LongTensor],
+        gt_ids: Sequence[LongTensor] = None,
```
DaloroAT, line 31: `Optional[Sequence[LongTensor]] = None`

oml/retrieval/retrieval_results.py
```diff
@@ -24,34 +26,46 @@
...
+            retrieved_ids: First gallery indices retrieved for every query with the size of ``n_query``.
+                Every index is within the range ``(0, n_gallery - 1)``.
+            gt_ids: Gallery indices relevant to every query with the size of ``n_query``.
+                Every element is within the range ``(0, n_gallery - 1)``
 
         """
-        assert distances.shape == retrieved_ids.shape
-        assert distances.ndim == 2
+        for d, r in zip(distances, retrieved_ids):
+            if not (d[:-1] <= d[1:]).all():
```
DaloroAT, line 43: Logically it's okay to have empty distances and retrieved_ids (e.g. after filtering by threshold) for some queries, and it's funny that all these checks pass 😁 I was not sure about how 1 check should work for empty tensors.

oml/retrieval/retrieval_results.py
```diff
@@ -64,14 +78,16 @@ def compute_from_embeddings(
...
+                gallery size if needed. Note, some queries may get less than this number of retrieved items if they
+                don't have enough gallery items available.
 
         """
         assert len(embeddings) == len(dataset), "Embeddings and dataset must have the same size."
 
         if SEQUENCE_COLUMN in dataset.extra_data:
-            sequence_ids = LongTensor(pd.factorize(dataset.extra_data[SEQUENCE_COLUMN])[0])
+            sequence = pd.Series(dataset.extra_data[SEQUENCE_COLUMN])
+            sequence_ids = LongTensor(pd.factorize(sequence)[0])
```
DaloroAT, line 90: Maybe it's not related to this PR, but I recommend to add param `sort=True` to `pd.factorize`. I worked in sim-api with categories or some handcrafted column and my `df` was recreated in runtime each time depends on params. And each time raw values encoded to different values (I used them later for debugging). Unique values of such "categories" were the same, just with different amounts and orders.

It's not happening if `df` remains the same, but why not ...

```python
import pandas as pd
from random import shuffle

value = [1, 1, 2, 3, 1, 2, 3, 4]
shuffle(value)

df = pd.DataFrame({"raw": value})
df["encode"] = pd.factorize(df["raw"])[0]

df = df.sort_values(by="raw")

print(df)
```

oml/retrieval/retrieval_results.py
```diff
@@ -89,17 +105,19 @@ def compute_from_embeddings(
         return RetrievalResults(distances=distances, retrieved_ids=retrieved_ids, gt_ids=gt_ids)
 
     def __str__(self) -> str:
+        max_el_to_show = 100
```
DaloroAT, line 108: class attribute?

oml/retrieval/retrieval_results.py
```diff
@@ -125,13 +143,18 @@ def visualize(
         if not isinstance(dataset, IQueryGalleryDataset):
             raise TypeError(f"Dataset has to support {IQueryGalleryDataset.__name__}. Got {type(dataset)}.")
 
+        nq1, nq2 = len(self.retrieved_ids), len(dataset.get_query_ids())
+        if nq1 != nq2:
+            raise RuntimeError(f"Number of queries in RetrievalResults and Dataset must match: {nq1} != {nq2}")
```
DaloroAT, line 148: Extract names of classes programmatically as above

DaloroAT: `type(dataset)` will show you full path to class, it's not friendly
`dataset.__class__.__name__` shows only name
and 
`self.__class__.__name__` for RetrievalResults

DaloroAT:
> - RetrievalResults uses Sequence of Tensors which may have different size. In other words, it allows us to support the case when queries have different number of retrieved items.
> - Consequently, changed batched_knn, retrieval_metrics and PairwiseReranker to support new input type.

image

oml/functional/knn.py
```diff
@@ -43,36 +43,41 @@ def batched_knn(
...
+            gt_ids.extend([row.nonzero().view(-1) for row in mask_gt_b])  # type: ignore
 
         distances_b[mask_to_ignore_b] = float("inf")
-        distances[i : i + bs, :], retrieved_ids[i : i + bs, :] = torch.topk(
-            distances_b, k=top_n, largest=False, sorted=True
-        )
+        distances_b_sorted, retrieved_ids_b = torch.topk(distances_b, k=top_n, largest=False, sorted=True)
+
+        # every query may have arbitrary number of retrieved items, so we are forced to use a loop to store the results
+        for dist, ids in zip(distances_b_sorted, retrieved_ids_b):
```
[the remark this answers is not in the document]
AlekseySh: I've check with TQDM that this function is not the bottleneck, so, let's do optimization later (after more urgent stuff)

oml/functional/metrics.py
```diff
@@ -294,42 +294,45 @@ def calc_precision(gt_tops: Tensor, n_gt: Tensor, top_k: Tuple[int, ...]) -> Lis
...
+        >>> n_gt = [2, 3, 5]
         >>> calc_precision(gt_tops, n_gt, top_k=(1, 2))
         [tensor([1., 0., 0.]), tensor([0.5000, 0.5000, 0.0000])]
+
     """
     check_if_nonempty_positive_integers(top_k, "top_k")
-    top_k = _clip_max_with_warning(top_k, gt_tops.shape[1])
+
+    def precision_single(is_correct: BoolTensor, n_gt_: int, k_: int) -> float:
+        k_ = min(k_, len(is_correct))
```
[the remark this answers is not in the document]
AlekseySh: You are right. But support of this case is added in the next PR: https://github.com/OML-Team/open-metric-learning/pull/566

oml/functional/metrics.py
```diff
@@ -294,42 +294,45 @@ def calc_precision(gt_tops: Tensor, n_gt: Tensor, top_k: Tuple[int, ...]) -> Lis
...
+        value = torch.cumsum(is_correct, dim=0)[k_ - 1] / min(n_gt_, k_)
+        return float(value)
```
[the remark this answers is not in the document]
AlekseySh: I return float from the inner function because I expect here just a single value without any extra dimensions which I may have if I keep it as tensor

oml/retrieval/postprocessors/pairwise.py
```diff
@@ -36,92 +35,92 @@ def __init__(
...
 
             # Note, if new and old distances are already sorted, we don't apply any scaling.
 
         """
-        assert retrieved_ids.shape == distances.shape
-        assert len(retrieved_ids) == len(dataset.get_query_ids())
-        assert retrieved_ids.shape[1] <= len(dataset.get_gallery_ids())
+        assert len(dataset.get_query_ids()) == len(
+            rr.retrieved_ids
+        ), "RetrievalResults and dataset must have the same number of queries."
```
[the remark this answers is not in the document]
AlekseySh: changed, thx

oml/retrieval/retrieval_results.py
AlekseySh: agree, done
also changed naming: `_max_elements_in_str_repr`

oml/retrieval/retrieval_results.py
AlekseySh: done, thx!

oml/retrieval/retrieval_results.py
AlekseySh: Nice! I didn't know about this. And thank you for the snippet. Changed.

PR #571, opened 2024-05-25
Added text support

AlekseySh:
Added:
* `TextBaseDataset`, `TextLabeledDataset`, `TextQueryGalleryLabeledDataset`, `TextQueryGalleryDataset`, `get_mock_texts_dataset`, `visualise_text`
* Added `HFWrapper` to wrap models from HuggingFace library
* `download_mock_dataset` -> `download_mock_dataset` (the original name is also kept for back compatibility)


AlekseySh:
I will add examples (code snippets) in the next PR.

oml/datasets/texts.py
```diff
@@ -0,0 +1,239 @@
...
+from oml.interfaces.datasets import (
+    IBaseDataset,
+    ILabeledDataset,
+    IQueryGalleryDataset,
+    IQueryGalleryLabeledDataset,
+    IVisualizableDataset,
+)
+from oml.utils.misc import visualise_text
+
+THuggingFaceTokenizer = Any
```
leoromanovich, line 30: May be we should not specify tokeriner as HF? Or use specific class from tokenizers (it already installed with hf). from tokenizers import Tokenizer

leoromanovich: Found that comment below. I see. May be add it here too?

`# We don't use actual types programmatically in order not to bring optional dependency here`

oml/datasets/texts.py
```diff
@@ -0,0 +1,239 @@
...
+        self._max_length = max_length
+
+        self.input_tensors_key = input_tensors_key
+        self.index_key = index_key
+
+    def __len__(self) -> int:
+        return len(self._texts)
+
+    def __getitem__(self, item: int) -> Dict[str, Any]:
+        encoded_text = self._tokenizer(
```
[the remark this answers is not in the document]
DaloroAT: To get help and highlighting from the IDE if the optional library is installed, you can use this snippet (`BoolTensor` as example):

```python
from typing import TYPE_CHECKING, Any


if TYPE_CHECKING:
    try:
        from torch import BoolTensor
    except ImportError:
        BoolTensor = Any
else:
    BoolTensor = Any
```

Or raise a new error in `except ...` block with recommendations to install transformers

it should work with IDE and mypy.

oml/utils/misc.py
```diff
@@ -162,6 +163,40 @@ def pad_array_right(arr: np.ndarray, required_len: int, val: Union[float, int])
...
+    margin = 10
+    max_width = im_size - 2 * margin
+    text_lines = []
+
+    # Split the text into lines that fit within the max_width
+    for line in text.split("\n"):
+        words = line.split()
+        current_line = []  # type: ignore
+        while words:
+            while words and draw.textlength(" ".join(current_line + [words[0]]), font=font) <= max_width:
```
DaloroAT, line 183: you enter an infinite loop if you can't fit your word into a line of pixels with the provided font

```python
plt.imshow(visualise_text("a" * 60, (150, 150, 150)))
plt.show()
```

I don't know how realistic such a situation is, but missing 1-2 spaces and getting a long single word is not impossible, especially for raw texts

DaloroAT: I just wanted to check how it looks and it freezes with long word

oml/const.py
```diff
@@ -71,6 +71,7 @@ def get_cache_folder() -> Path:
 CROP_KEY = "crop"  # the format is [x1, y1, x2, y2]
 
 # Required dataset format:
```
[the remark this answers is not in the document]
AlekseySh: You are right, done

oml/utils/misc.py
AlekseySh: you are right

I tried to cover all corner cases in the current function but gave up.

Finally, I reworked it using matplotlib and added a few tests on huge inputs.

oml/datasets/texts.py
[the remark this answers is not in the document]
AlekseySh: okay, updated types here and for HFWrapper

oml/datasets/texts.py
AlekseySh: okay, updated types here and for HFWrapper

oml/datasets/texts.py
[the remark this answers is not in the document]
AlekseySh: I've spent plenty of time trying to make work this method but finally gave up. (Type hints did not work)

I've added runtime type checking instead. Here, and in `HFWrapper`

oml/datasets/texts.py
AlekseySh: I've added runtime type checking

Issue #583, opened 2024-06-07
RetrievalResults when separated query gallery 


AlekseySh:
Here is the current usage of `RetrievalResults`:

```python
dataset = QueryGalleryDataset(df_test)
embeddings = inference(extractor, dataset)

rr = RetrievalResults.from_embeddings(embeddings, dataset)
```

The problem may appear in a few following cases:
* Queries and galleries have different nature: for examples queries are texts, galleries are images, and a model is something like CLIP having image and text encoders.
* Gallery embeddings are pre-computed, but we receive new batches of queries online.

To address this cases, we may implement something like:
```python
dataset_g = BaseDataset(galleries)
dataset_q = BaseDataset(queries) 

embeddings_g = inference(model_g, dataset_g)
embeddings_q = inference(model_q, dataset_q)

rr = RetrievalResults.from_qg_embeddings(embeddings_q, embeddings_g, dataset_q, dataset_g)

```

In the example above different nature of queries and galleries is acceptable. We can also easily organize loop over queries without need of recalculating gallery embeddings.

AlekseySh:
- [ ] all checked

PR #584, opened 2024-06-09
Supporting separated queries and galleries

AlekseySh:
The motivation: https://github.com/OML-Team/open-metric-learning/issues/583

Changelog:
* Added `batched_knn_qg`. So `batched_knn` is just an adapter for `batched_knn_qg`.
* Added a few methods to `RetrievalResults`: `from_embeddings_qg`, `visuaze_qg`, `visualize_with_functions`. Also added these methods to docs.
* Added an example to Readme and docs where new methods are shown.
* Updated tests: added text modality to `test_retrieval_results`, added new `test_retrieval_results_separated_qg`


oml/functional/knn.py
```diff
@@ -1,55 +1,59 @@
...
+    labels_gallery: Optional[LongTensor] = None,
+    ids_query: Optional[LongTensor] = None,
+    ids_gallery: Optional[LongTensor] = None,
+    sequence_ids_query: Optional[LongTensor] = None,
+    sequence_ids_gallery: Optional[LongTensor] = None,
     bs: int = BS_KNN,
     verbose: bool = False,
 ) -> Tuple[Sequence[FloatTensor], Sequence[LongTensor], Optional[Sequence[LongTensor]]]:
     """
+    Arguments have the same meaning as in `batched_knn`.
```
leoromanovich, line 33: I believe here we need to describe arguments anyway. We shouldn't expect from user to know api of different method.

leoromanovich: or mark it with underscore as a private method

AlekseySh: discussed offline: it's okay for internal function

PR #596, opened 2024-06-14
Extra dependances

ayasyrev:
- [* ] I've checked contribution [guide](https://open-metric-learning.readthedocs.io/en/latest/oml/contributing.html).

Add "extra" options for pip install.
- tests
- optional
- loggers
- nlp

If OK, add info to documentation.
Now we can move "tensorboard" to loggers and create cv requrements file

### DIFF
--- a/ci/requirements_optional.txt
+++ b/ci/requirements_optional.txt
@@ -1,2 +1,0 @@
-grad-cam==1.4.6
-jupyter>=1.0.0
--- a/ci/requirements_optional.txt
+++ b/ci/requirements_optional.txt
@@ -1,0 +1,6 @@
+neptune>=1.0.0, <1.10.1
+wandb>=0.15.4
+mlflow>=2.0.0
+clearml>=1.5.0
+tensorboard
+grad-cam==1.4.6
--- a/ci/requirements_optional.txt
+++ b/ci/requirements_optional.txt
@@ -2,5 +2,4 @@ neptune>=1.0.0, <1.10.1
 wandb>=0.15.4
 mlflow>=2.0.0
 clearml>=1.5.0
-tensorboard
 grad-cam==1.4.6
--- a/ci/requirements_optional.txt
+++ b/ci/requirements_optional.txt
@@ -1,7 +1,2 @@
 grad-cam==1.4.6
 jupyter>=1.0.0
-neptune>=1.0.0, <1.10.1
-wandb>=0.15.4
-mlflow>=2.0.0
-clearml>=1.5.0
-transformers>=4.0
--- a/ci/requirements_tests.txt
+++ b/ci/requirements_tests.txt
@@ -1,9 +1,3 @@
 pytest>=7.2
 pre-commit==2.15.0
 python-dotenv>=0.17.0
-neptune>=1.0.0, <1.10.1
-wandb>=0.15.4
-mlflow>=2.0.0
-clearml>=1.5.0
-tensorboard
-grad-cam==1.4.6
--- a/ci/requirements_tests.txt
+++ b/ci/requirements_tests.txt
@@ -1,3 +1,8 @@
 pytest>=7.2
 pre-commit==2.15.0
 python-dotenv>=0.17.0
+neptune>=1.0.0, <1.10.1
+wandb>=0.15.4
+mlflow>=2.0.0
+clearml>=1.5.0
+tensorboard
--- a/ci/requirements_tests.txt
+++ b/ci/requirements_tests.txt
@@ -6,3 +6,4 @@ wandb>=0.15.4
 mlflow>=2.0.0
 clearml>=1.5.0
 tensorboard
+grad-cam==1.4.6
--- a/setup.py
+++ b/setup.py
@@ -18,8 +18,6 @@ def load_version() -> str:
         return re.search(r'^__version__ = [\'"]([^\'"]*)[\'"]', f.read(), re.M).group(1)
 
 
-TESTS_REQUIRE = load_requirements("ci/requirements_tests.txt")
-OPTIONAL_REQUIRE = load_requirements("ci/requirements_optional.txt")
 NLP_REQUIRE = load_requirements("ci/requirements_nlp.txt")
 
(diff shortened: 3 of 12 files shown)


ayasyrev:
Now "extra"  is: "optional", "nlp", "full" and "tests".
Later will be "cv", "pipelines", "audio".

ci/requirements_optional.txt
```diff
@@ -1,7 +1,2 @@
 grad-cam==1.4.6
```
AlekseySh, line 1: General idea for optional: **let's remove it.**

1. Let's remove jupyer because we don't use it anymore. We had this requirement to test a notebook that has already been removed from the library.

2. After that we have the only optional req -- gradcam. I see two options. First, we can move it into `cv` tag. The problem here is gradcam has some extra [dependencies](https://github.com/jacobgil/pytorch-grad-cam/blob/master/requirements.txt), but at the same time it's value is relatively small and I don't think all `cv` users really not to deal with installing gradcam. Thus, I suggest the second option: move gradcam to the `tests`. So, the users will install this library in those rare cases where they need it (similar to loggers logic).

What do u think?

ci/requirements_tests.txt
```diff
@@ -1,3 +1,8 @@
 pytest>=7.2
 pre-commit==2.15.0
 python-dotenv>=0.17.0
+neptune>=1.0.0, <1.10.1
+wandb>=0.15.4
+mlflow>=2.0.0
+clearml>=1.5.0
+tensorboard
```
AlekseySh, line 8: if you agree with the comment above, we add `gradcam` here

AlekseySh (changes requested):
I suggest some changes. If we go this way, we will only have "full" and "nlp" pip tags for now. We will add "audio" and "cv" later, but not "optional" and "tests"

Also, let's update installation instruction. Please, check contribution guide to understand how to change something in Readme.

ci/requirements_optional.txt
ayasyrev: 1 - agreed. 
2 - IMHO gradcam not for test. it is for cv. And yes - it has a lot of dependencies, especially scikit-learn. Lets move it to test, maybe later will be better solution.

AlekseySh: 2 - Agree with you, @ayasyrev . Yep, we need loggers and gradcam in tests, but it's file naming is a bit weir. We may have `requirements_optional.txt`, where we put gradcam and loggers, but we don't make `optional` tag for cv. Instead, we just install this requirements in test time. As for the `requirements_test.txt`, we just keep there pytest and linters, which is more semantically correct in terms of naming. What do u think?

ayasyrev: OK, test - for test related stuff.

AlekseySh:
@ayasyrev everything looks okay to me

let's finish with my comment above
and fix that numpy error, it should be simple

after that we are good to merge

DaloroAT:
Do we need to add typings from transformers for NLP related  modules (datasets, models)? Right now placeholders with Any type there. Or we can do it in separate PR.

AlekseySh:
@DaloroAT 

https://github.com/OML-Team/open-metric-learning/pull/596#issuecomment-2171833135

I've tried and failed in the previous PRs. Spent a few hours on that. Seems like I could not make work type hints when they created in if-else condition. Probably I can may it work... I dont'know

AlekseySh (changes requested):

DaloroAT:
> @DaloroAT
> 
> [#596 (comment)](https://github.com/OML-Team/open-metric-learning/pull/596#issuecomment-2171833135)
> 
> I've tried and failed in the previous PRs. Spent a few hours on that. Seems like I could not make work type hints when they created in if-else condition. Probably I can may it work... I dont'know

DaloroAT:
> I've tried and failed in the previous PRs. Spent a few hours on that. Seems like I could not make work type hints when they created in if-else condition. Probably I can may it work... I dont'know

Now you can import Tokenizers and models from transformers in codebase without if-else. It makes sense, because lib might be installed with `[nlp]` extra

AlekseySh:
> Now you can import Tokenizers and models from transformers in codebase without if-else. It makes sense, because lib might be installed with [nlp] extra

@DaloroAT 
what if I installed library without `nlp`? I will get error when imports chain will be executed

setup.py
```diff
@@ -18,12 +18,19 @@ def load_version() -> str:
...
+
+
 setup(
     # technical things
     version=load_version(),
     packages=find_packages(exclude=["ci", "docs", "pipelines", "tests*"]),
     python_requires=">=3.8,<4.0",
     install_requires=load_requirements("ci/requirements.txt"),
+    extras_require={
+        "nlp": NLP_REQUIRE,
```
DaloroAT, line 31: @AlekseySh 
Let's discuss here in some "thread"

>what if I installed library without nlp? I will get error when imports chain will be executed

Why do you need to import nlp related parts without nlp related dependencies? 

Or more general question: will `Any` typing also be used in the CV and audio domains for transforms and models?

DaloroAT: I think, proper solution - import directly from dependencies and if dependencies are not installed - show message to install them or install oml with extra (oml[nlp], oml[cv], oml[audio]).

We can add some helpers and call them before any imports. Like 

```
verify_cv_installed()
import os

import cv2
```

ayasyrev: I will do it on next step - separate cv dep's.

AlekseySh: > Why do you need to import nlp related parts without nlp related dependencies?

because we have some import in `__init__.py`, that is why an error may be triggered 

generally you are right, we can rework it in the next PRs

DaloroAT: Sure. CV is just an example to "think about". My whole point is not about `nlp` itself, but about any domain

AlekseySh: this PR is about requirement files and pip tags, let's keep it simple and rework python code in the next ones

ayasyrev:
Return tensorboard to main requirements. It needs for short tests.

PR #597, opened 2024-06-16
Fix freeze and errors in `tests/test_oml/test_ddp`

DaloroAT:
- [x] I've checked contribution [guide](https://open-metric-learning.readthedocs.io/en/latest/oml/contributing.html).

There were problems that DDP tests in `tests/test_oml/test_ddp` would either freeze or crash with an error. 

I made the following changes:
1) The file for communication between ranks is formed based on the test case (name of the function being tested, Python version, a hash of case argument values), which is unique and known to all ranks. Previously, the file had the same name and we hoped it would be deleted in time or unused.
2) Increased timeout for DDP communication. The previous 10 seconds may not be enough for DDP tests on a matrix of 4 versions of pythons.
3) I discovered that if you run only DDP tests, they almost always freeze. Most likely, CI machines on GitHub cannot handle such a load. This does not happen if you force the number of threads to be 1 for `torch`.

Runs with DDP only after changes (`pytest -sv tests/test_oml/test_ddp` hardcoded for `short_tests`):
- https://github.com/OML-Team/open-metric-learning/actions/runs/9626871872/job/26553326473
- https://github.com/OML-Team/open-metric-learning/actions/runs/9626951865/job/26553495699
- https://github.com/OML-Team/open-metric-learning/actions/runs/9627028721/job/26553666252


AlekseySh (approved):
Amazing! The problem was sooo annoying! Thank you :)

Issue #582, opened 2024-06-07
Support audio


AlekseySh:
Let's add audio to the already supporting modalities like images and texts.

**The plan:**
* Add an analogue of `get_mock_images_dataset`
* Add the corresponding datasets inherited from the corresponding interfaces
* Add a sound processing NN model to Zoo, or add a wrapper to 3rd party sound models (like `HFWrapper`)
* Based on the objects above, create `train` / `val` example.
* Evaluate an example on a dataset having larger scale in order to check the example may be applied on real data.

amanteur:
Hi, I'll do it!

AlekseySh:
@amanteur What a coincidence! 

You are welcome! :)

PR #598, opened 2024-06-17
Add audio datasets

amanteur:
- [*] I've checked contribution [guide](https://open-metric-learning.readthedocs.io/en/latest/oml/contributing.html).

This PR introduces support for audio datasets as a new modality, addressing part of the feature request outlined in [issue #582](https://github.com/OML-Team/open-metric-learning/issues/582). Below are the key additions:

- **Mock Audio Dataset**: Added the `get_mock_audios_dataset` function, which mirrors the functionality of `get_mock_images_dataset`. This function provides a subset of the [VoxCeleb1 dataset](https://www.robots.ox.ac.uk/~vgg/data/voxceleb/vox1.html), commonly used for voice recognition and verification tasks.

- **Dataset Classes**: Introduced new dataset classes to facilitate audio data handling: `AudioBaseDataset, AudioLabeledDataset, AudioQueryGalleryDataset, AudioQueryGalleryLabeledDataset`
These classes are analogous to the existing image dataset classes, enabling similar operations for audio data.

- **Audio Visualization**: Implemented functions to visualize audio data both as standalone images and within HTML code, including embedded audio playback.


oml/datasets/audios.py
```diff
@@ -0,0 +1,353 @@
...
+            extra_data=extra_data,
+            sr=sr,
+            num_seconds=num_seconds,
+            num_channels=num_channels,
+            input_tensors_key=input_tensors_key,
+            input_frame_offset_key=input_frame_offset_key,
+            index_key=index_key,
+        )
+
+    def get_query_ids(self) -> LongTensor:
```
AlekseySh, line 311: please, don't copy my mistake :) 

take a look at the [fix](https://github.com/OML-Team/open-metric-learning/commit/0e3776f8507bb3ed95bcdbd7ebfdb3515e724361)

oml/utils/download_mock_dataset.py
```diff
@@ -188,4 +192,56 @@ def get_mock_texts_dataset() -> Tuple[pd.DataFrame, pd.DataFrame]:
     return df_train, df_val
 
 
-__all__ = ["get_mock_images_dataset", "get_mock_texts_dataset"]
+def get_mock_audios_dataset(
```
AlekseySh, line 195: seems like there is almost no difference between `get_mock_images_dataset` and `get_mock_texts_dataset`

let's introduce a function `_download_dataset` implementing shared logic and reuse it in both `get_mock_images_dataset` and  `get_mock_texts_dataset`?

oml/utils/audios.py
```diff
@@ -0,0 +1,86 @@
...
+
+    fig.canvas.draw()
+    width, height = fig.canvas.get_width_height()
+    image = np.frombuffer(fig.canvas.tostring_rgb(), dtype=np.uint8)
+    image = image.reshape(height, width, 3)
+    plt.close(fig)
+    return image
+
+
+def visualize_audio_html(audio: FloatTensor, spec_repr: FloatTensor, sr: int, color: str = "black") -> str:
```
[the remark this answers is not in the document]
AlekseySh: the same about another similar method: `def visualize_html`

oml/datasets/audios.py
```diff
@@ -0,0 +1,353 @@
...
+    def __init__(
+        self,
+        paths: List[Union[str, Path]],
+        dataset_root: Optional[Union[str, Path]] = None,
+        extra_data: Optional[Dict[str, Any]] = None,
+        sr: int = DEFAULT_SAMPLE_RATE,
+        num_seconds: float = DEFAULT_DURATION,
+        num_channels: int = DEFAULT_NUM_CHANNELS,
+        input_tensors_key: str = INPUT_TENSORS_KEY,
+        input_frame_offset_key: str = INPUT_FRAME_OFFSET_KEY,
```
AlekseySh, line 56: I'm not sure we need this key and suggest to remove the key and the constant `INPUT_FRAME_OFFSET_KEY`. Modality related information should not appear outside of the dataset.

The purpose of `extra_data` is to store some external information provided from outside (usually by user in runtime or when we have extra columns in the dataframe). To be honest, I did not understand from where we take this offsets, because there are no offsets in the dataframe.

By the way? Do you need to be used later, for example, in the model?

oml/datasets/audios.py
```diff
@@ -0,0 +1,353 @@
...
+
+    """
+
+    def __init__(
+        self,
+        paths: List[Union[str, Path]],
+        dataset_root: Optional[Union[str, Path]] = None,
+        extra_data: Optional[Dict[str, Any]] = None,
+        sr: int = DEFAULT_SAMPLE_RATE,
+        num_seconds: float = DEFAULT_DURATION,
```
AlekseySh, line 53: should we name it `max_num_seconds`?

oml/datasets/audios.py
```diff
@@ -0,0 +1,353 @@
...
+        Load and process an audio file.
+
+        Args:
+            item (int): Dataset item index.
+
+        Returns:
+            FloatTensor: Processed audio tensor.
+        """
+        path = self._paths[item]
+        assert (
```
AlekseySh, line 142: probably it's too late to check an extension, could we fail earlier? in init?

oml/datasets/audios.py
```diff
@@ -0,0 +1,353 @@
...
+            path.suffix in AUDIO_EXTENSIONS
+        ), f"Input audio has invalid extension '{path.suffix}'. It should be one of {AUDIO_EXTENSIONS}."
+        audio, sr = torchaudio.load(path)
+        frame_offset = None if self.frame_offsets is None else self.frame_offsets[item]
+        audio = self._downmix_and_resample(audio, sr)
+        audio = self._trim_or_pad(audio, frame_offset)
+        return audio
+
+    def get_spectral_repr(self, audio: FloatTensor, params: Dict[str, Any] = DEFAULT_MELSPEC_PARAMS) -> FloatTensor:
```
AlekseySh, line 151: do we need this method as public as far as we have two `visualize` methods which look very similar?
moreover, this method is almost static which may be confusing

i suggest to make the method private

tests/test_oml/test_datasets/test_datasets.py
```diff
@@ -13,6 +13,10 @@
     TEXTS_COLUMN,
```
AlekseySh, line 13: Is that possible to make one more test file under `.../test_datasets/test_audio.py` and check audio specific stuff?
We can try to instantiate dataset items with different values of `sr`, `num_seconds`, offsets and perform other checks that make sense to you.

oml/const.py
```diff
@@ -104,3 +104,16 @@ def get_cache_folder() -> Path:
...
 HYDRA_BEHAVIOUR = "1.1"
+
+# Audio specific constants
+MOCK_AUDIO_DATASET_PATH = CACHE_PATH / "mock_audio_dataset"
+MOCK_AUDIO_DATASET_URL_GDRIVE = "https://drive.google.com/drive/folders/1aYqXBNnERFRIzxr_6cBVO_v_qz2dU7ta"
+MOCK_AUDIO_DATASET_MD5 = "87ba6367ac1231c6be4ccca5e8ace837"
+MOCK_AUDIO_DATASET_CSV_NAME = "df.csv"
+
+AUDIO_EXTENSIONS = [".mp3", ".wav", ".ogg", ".flac"]
+INPUT_FRAME_OFFSET_KEY = "frame_offset"
```
AlekseySh, line 115: Note sure we need this constant, see comments below

AlekseySh (changes requested):
Hey, and thank you for your great work! I like how similar to other modalities these datasets look like :)

AlekseySh:
@amanteur please make any non meaningful PR so CI triggers on your comments automatically

oml/datasets/audios.py
amanteur: the idea is to have this method as public, so a user could override it with different spectral representation extraction method (e.g. spectrogram/CQT/MFCC/...) for visualisation

oml/datasets/audios.py
amanteur: i thought a lot on this question, so:
1. I changed naming `frame_offset -> start_time`, so it would represent not the number of frames, but the number of seconds.
2. Instead of adding explicitly `start_time`(prev. `frame_offset`) column as dataset argument, I added bool arg `use_random_start`, which works as following:
  - if True, there will be random starts and no `start_time` will be applied
  - If False, it will search for `start_time` in `extra_data`, and if it fails, there will be Exception.

ci/requirements_optional.txt
```diff
@@ -5,3 +5,4 @@ wandb>=0.15.4
 mlflow>=2.0.0
 clearml>=1.5.0
 transformers>=4.0
+torchaudio>=2.0,<=2.1.1
```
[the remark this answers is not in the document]
amanteur: done

oml/const.py
```diff
@@ -104,3 +104,16 @@ def get_cache_folder() -> Path:
...
+DEFAULT_DURATION = 3.0
+DEFAULT_SAMPLE_RATE = 16_000
+DEFAULT_NUM_CHANNELS = 1
```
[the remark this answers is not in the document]
amanteur: done

oml/datasets/audios.py
```diff
@@ -0,0 +1,353 @@
...
+            dataset_root=dataset_root,
+            extra_data=extra_data,
+            sr=sr,
+            num_seconds=num_seconds,
+            num_channels=num_channels,
+            input_tensors_key=input_tensors_key,
+            input_frame_offset_key=input_frame_offset_key,
+            index_key=index_key,
+            labels_key=labels_key,
+        )
```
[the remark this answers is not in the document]
amanteur: done

oml/utils/audios.py
```diff
@@ -0,0 +1,86 @@
...
+
+from oml.const import BLACK, TColor
+
+
+def visualize_audio(spec_repr: FloatTensor, color: TColor = BLACK, draw_bbox: bool = True) -> NDArray[np.uint8]:
+    """
+    Visualizes an audio spectrogram representation as an image with optional colored bounding box.
+
+    Parameters:
+        spec_repr (FloatTensor): The spectral representation data as a 2D torch tensor.
```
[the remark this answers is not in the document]
amanteur: done

oml/utils/audios.py
```diff
@@ -0,0 +1,86 @@
...
+        audio (FloatTensor): The raw audio data as a 1D torch tensor.
+        spec_repr (FloatTensor): The spectral representation data as a 2D torch tensor.
+        sr (int): The sample rate of the audio data.
+        color (str): The color used for the HTML border around the spectral representation and audio player.
+
+    Returns:
+        str: A string containing HTML that embeds both the image and the audio player.
+    """
+    # generate base64 image
+    fig, ax = plt.subplots(figsize=(2.56, 2.56), dpi=100)
```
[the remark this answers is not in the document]
amanteur: done

oml/utils/audios.py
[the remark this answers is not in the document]
amanteur: done

oml/utils/download_mock_dataset.py
amanteur: done

oml/datasets/audios.py
amanteur: done

oml/datasets/audios.py
amanteur: done

tests/test_oml/test_datasets/test_datasets.py
amanteur: done

oml/const.py
amanteur: done

AlekseySh (commented):
Thank you for addressing the comments!

Sometimes you use "Args", but sometimes "Params". Please, follow the existing library's code style. I think CI/CD should catch it when activated on your comments.

oml/datasets/audios.py
AlekseySh: Hm, what is the general purpose of random start? Is that type of training augmentation?
Do I understand correctly, that we never use such option in the validation, because we want to have reproducible results?

tests/test_oml/test_datasets/test_datasets.py
```diff
@@ -143,3 +152,28 @@ def test_image_datasets() -> None:
     check_base(dataset_qgl)
     check_query_gallery(dataset_qgl, df_val)
     check_labeled(dataset_qgl, df_val)
+
+
+def test_audio_datasets() -> None:
```
AlekseySh, line 157: pls, use `@pytest.mark.needs_optional_dependency` if you need torchaudio for this test

tests/test_oml/test_datasets/test_audio.py
```diff
@@ -0,0 +1,49 @@
+import random
```
AlekseySh, line 1: pls, use `@pytest.mark.needs_optional_dependency` if you need torchaudio for some tests below

oml/datasets/audios.py
amanteur: yeah, it is basically some kind of augmentation and yeah, we'll never use it in the validation. So there can be several situations:
- user defined start times in csv/extra data, so these values will be used both on training and validation
- user did not defined offsets in csv/extra data, so on the training section audio may have random start or start on beginning (0) and on validation only on beginning too (I'll add this logic).

Essentially, we need these start_times/random_start because audios are mostly longer than it is required for the model, so we need to decide how to trim it. A lot of datasets come with the `start_time` of the training/validation segment, so you just cut it from the particular point, but sometimes they dont, so we need to trim them mostly randomly.

oml/datasets/audios.py
amanteur: Suppose, I train and validate a model with [CQTs](https://en.wikipedia.org/wiki/Constant-Q_transform) (which looks a bit different from melspectrograms), and when validating via retrieval I see melspectrograms instead of CQTs and can not gain insights why model works in such way. 
So, I need to rewrite the visualisation method. 

And why then do we need the private method if it won't be (potentially) used?

oml/utils/audios.py
```diff
@@ -0,0 +1,107 @@
...
+    plt.savefig(buf, format="png", bbox_inches="tight")
+    buf.seek(0)
+    img: Union[NDArray[np.uint8], str] = (
+        base64.b64encode(buf.getvalue()).decode("ascii") if return_b64 else np.array(Image.open(buf), dtype=np.uint8)
+    )
+    plt.close(fig)
+    return img
+
+
+def visualize_audio(spec_repr: FloatTensor, color: TColor = BLACK, draw_bbox: bool = True) -> NDArray[np.uint8]:
```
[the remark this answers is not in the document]
amanteur: `numpy 1.21` 
[link](https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray)

oml/datasets/audios.py
amanteur: As an idea, I can remove this method from class and move it to `utils`. Then, I'll add the additional argument (e.g. `spec_repr_func: Callable[[FloatTensor], FloatTensor]`) in `visualize/visualize_with_player` methods.

something like this:
`datasets/audios`
```
from utils.audios import default_spec_repr_func
class Dataset(...):
...
    def visualize(
        self,
        item: int,
        color: TColor = BLACK,
        spec_repr_func: Callable[[FloatTensor], FloatTensor] = default_spec_repr_func,
    ) -> NDArray[np.uint8]:
        """
        Visualize an audio file.

        Args:
            item: Dataset item index.
            color: Color of the plot.
            spec_repr_func: Spectral representation extraction function

        Returns:
            Array representing the image of the plot.
        """
        audio = self.get_audio(item)
        spec_repr = spec_repr_func(audio).squeeze(0)
        return visualize_audio(spec_repr=spec_repr, color=color)
...
``` 
`utils/audios`
```
def default_spec_repr_func(audio: FloatTensor) -> FloatTensor:
    """
    Generate a log-scaled Spectrogram from an audio signal.
    Used primarily for visualization.

    Args:
        audio: The input audio tensor.

    Returns:
        The spectral representation of the input audio tensor.
    """
    spec = Spectrogram(n_fft=1024)(audio)
    log_spec = torch.log1p(spec).squeeze(0)
    return log_spec
```

oml/const.py
```diff
@@ -40,6 +40,7 @@ def get_cache_folder() -> Path:
 STORAGE_CKPTS = STORAGE_URL + "/download/checkpoints"
 
 MOCK_DATASET_PATH = CACHE_PATH / "mock_dataset"
+MOCK_DATASET_CSV_NAME = "df.csv"
```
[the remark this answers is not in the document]
amanteur: done

oml/datasets/audios.py
```diff
@@ -0,0 +1,400 @@
...
+    IVisualizableDataset,
+)
+from oml.utils.audios import visualize_audio, visualize_audio_with_player
+
+
+class AudioBaseDataset(IBaseDataset, IVisualizableDataset):
+    """
+    The base class that handles audio specific logic.
+
+    Args:
```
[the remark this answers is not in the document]
amanteur: done

oml/datasets/audios.py
```diff
@@ -0,0 +1,400 @@
...
+        return visualize_audio_with_player(audio=audio, spec_repr=spec_repr, sr=self.sr, color=color)
+
+
+class AudioLabeledDataset(AudioBaseDataset, ILabeledDataset):
+    """
+    The dataset of audios having their ground truth labels.
+
+    """
+
+    def __init__(
```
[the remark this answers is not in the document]
amanteur: done

oml/datasets/audios.py
```diff
@@ -0,0 +1,400 @@
...
+
+class AudioQueryGalleryLabeledDataset(AudioLabeledDataset, IQueryGalleryLabeledDataset):
+    """
+    The **annotated** dataset of audios having `query`/`gallery` split.
+    To perform `1 vs rest` validation, where a query is evaluated versus the whole validation dataset
+    (except for this exact query), you should mark the item as ``is_query == True`` and ``is_gallery == True``.
+
+    """
+
+    def __init__(
```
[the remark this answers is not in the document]
amanteur: done

oml/datasets/audios.py
```diff
@@ -0,0 +1,400 @@
...
+
+class AudioQueryGalleryDataset(IVisualizableDataset, IQueryGalleryDataset):
+    """
+    The **non-annotated** dataset of audios having `query`/`gallery` split.
+    To perform `1 vs rest` validation, where a query is evaluated versus the whole validation dataset
+    (except for this exact query), you should mark the item as ``is_query == True`` and ``is_gallery == True``.
+
+    """
+
+    def __init__(
```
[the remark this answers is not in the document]
amanteur: done

oml/utils/audios.py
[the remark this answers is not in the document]
amanteur: done

tests/test_oml/test_datasets/test_datasets.py
amanteur: done

tests/test_oml/test_datasets/test_audio.py
amanteur: done

AlekseySh:
@amanteur Hey, if you have any problems with making CI green just let me know :)

oml/utils/audios.py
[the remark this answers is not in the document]
AlekseySh: I've [found](https://github.com/numpy/numpy/releases/tag/v1.21.0) the date of the release: it's June 2021. Hm, probably some users still use it. Anyway, in the rest of the repo we just use `np.ndarray`, let's do the same here in order to avoid any problems.

oml/datasets/audios.py
AlekseySh: Okay, let's do this way: `spec_repr_func: Callable[[FloatTensor], FloatTensor] = default_spec_repr_func,`

oml/datasets/audios.py
AlekseySh: Got it.

If there is a common practice when people make `random_start` the part of training transforms (augmentations)? There is an image related analogue: cropping images during the training (CropResize). If so, could we also move this to the augmentations?

As for the validation, it's look fine to me.

I'm also worried about potential conflict between `start_times` and `use_random_start_time`. Let's raise an error if 
`use_random_start_time=True`, but we also had `START_TIME_COLUMN` in our dataframe.

AlekseySh: Another thing: we should add and example with `START_TIME_COLUMN` to the mock dataset, as we did it for images: `df_with_bboxes.csv`. First, it will serve as example, second, we can use it in tests.

oml/const.py
```diff
@@ -101,6 +102,19 @@ def get_cache_folder() -> Path:
...
 IMAGE_EXTENSIONS = ["jpg", "jpeg", "JPG", "JPEG", "png"]
 
-# hydra provides ability to set its behaviour for convinient backwards compatibility
+# hydra provides ability to set its behaviour for convenient backwards compatibility
 # no matter what the current version is
 HYDRA_BEHAVIOUR = "1.1"
+
+# Audio specific constants
+MOCK_AUDIO_DATASET_PATH = CACHE_PATH / "mock_audio_dataset"
+MOCK_AUDIO_DATASET_URL_GDRIVE = "https://drive.google.com/drive/folders/1aYqXBNnERFRIzxr_6cBVO_v_qz2dU7ta"
```
AlekseySh, line 111: please, move to: `https://drive.google.com/drive/folders/1NcKnyXqDyyYARrDETmhJcTTXegO3W0Ju?usp  
 =   sharing`

AlekseySh:
@amanteur As we discussed on call, let's move optional imports into the functions bodies

oml/utils/audios.py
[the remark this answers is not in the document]
amanteur: what should I do if mypy goes angry on `np.ndarray`? 
```
 Missing type parameters for generic type "ndarray"
```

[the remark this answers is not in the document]
AlekseySh: probably you have a newer mypy, for CI's one it should be fine 
so just ignore that

oml/const.py
amanteur: done

oml/utils/audios.py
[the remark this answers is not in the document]
amanteur: done

oml/datasets/audios.py
amanteur: done

oml/datasets/audios.py
amanteur: done

amanteur:
i guess thats it

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+        extra_data: Optional[Dict[str, Any]] = None,
+        input_tensors_key: str = INPUT_TENSORS_KEY,
+        index_key: str = INDEX_KEY,
+        sr: int = DEFAULT_SAMPLE_RATE,
+        max_num_seconds: float = DEFAULT_DURATION,
+        num_channels: int = DEFAULT_AUDIO_NUM_CHANNELS,
+        spec_repr_func: Callable[[FloatTensor], FloatTensor] = default_spec_repr_func,
+    ):
+        """
+        Initializes the AudioQueryGalleryDataset.
```
AlekseySh, line 291: rm `Initializes the AudioQueryGalleryDataset.`

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+        input_tensors_key: str = INPUT_TENSORS_KEY,
+        index_key: str = INDEX_KEY,
+        labels_key: str = LABELS_KEY,
+        sr: int = DEFAULT_SAMPLE_RATE,
+        max_num_seconds: float = DEFAULT_DURATION,
+        num_channels: int = DEFAULT_AUDIO_NUM_CHANNELS,
+        spec_repr_func: Callable[[FloatTensor], FloatTensor] = default_spec_repr_func,
+    ):
+        """
+        Initializes the AudioQueryGalleryLabeledDataset.
```
AlekseySh, line 349: rm `Initializes the AudioQueryGalleryLabeledDataset.`

oml/utils/audios.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Args:
+        audio: The input audio tensor.
+
+    Returns:
+        The spectral representation of the input audio tensor.
+    """
+    from torchaudio.transforms import MelSpectrogram
+
+    melspectrogram = MelSpectrogram(sample_rate=16000, n_fft=2048, hop_length=512, n_mels=128)
```
AlekseySh, line 27: Should this `sample_rate=16_000` somehow be synchronized with ours used for model training?

oml/utils/audios.py
```diff
@@ -0,0 +1,127 @@
...
+    ax.imshow(spec_repr, aspect="auto", origin="lower")
+
+    # bbox and axes
+    if draw_bbox:
+        frame_thickness = 5
+        for axis in ["top", "bottom", "left", "right"]:
+            ax.spines[axis].set_linewidth(frame_thickness)
+            ax.spines[axis].set_edgecolor(
+                [c / 255 for c in color]
+            )  # TODO: change all colors from arrays to strings in vis functions
```
AlekseySh, line 61: not sure about this todo
could you explain?

tests/test_oml/test_datasets/test_audio.py
```diff
@@ -0,0 +1,95 @@
...
+
+@pytest.mark.needs_optional_dependency
+@pytest.mark.parametrize("num_channels", [1, 2])
+def test_downmix(df: pd.DataFrame, num_channels: int) -> None:
+    from oml.datasets.audios import AudioBaseDataset
+
+    dataset = AudioBaseDataset(df[PATHS_COLUMN].tolist(), num_channels=num_channels)
+    for item in dataset:
+        audio = item[dataset.input_tensors_key]
+        assert audio.shape[0] <= num_channels, f"Audio channels {audio.shape[0]} exceed specified {num_channels}"
```
AlekseySh, line 29: why `<=` but not a `!=` ?

tests/test_oml/test_datasets/test_audio.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@pytest.mark.needs_optional_dependency
+@pytest.mark.parametrize("sr", [8000, 16000, 22050, 44100, 48000])
+@pytest.mark.parametrize("max_num_seconds", [0.01, 0.5, 1.0, 3.0, 10.0, 100.0])
+def test_resample_trim_pad(df: pd.DataFrame, sr: int, max_num_seconds: float) -> None:
```
AlekseySh, line 35: the test looks good but let's remove a few parametrization parameters from each line in order not to waste time

tests/test_oml/test_datasets/test_audio.py
```diff
@@ -0,0 +1,95 @@
...
+            max_num_seconds * sr
+        ), f"Audio length {audio.shape[1]} does not match expected {int(max_num_seconds * sr)}"
+
+
+@pytest.mark.needs_optional_dependency
+def test_start_times(df: pd.DataFrame) -> None:
+    from oml.datasets.audios import AudioBaseDataset, AudioLabeledDataset
+
+    # test base dataset
+    start_times = [random.uniform(0, 1) for _ in range(len(df))]
```
AlekseySh, line 51: 1. the value can be greater than 1 because it's number of seconds
2. let's add a few empty values as well as I suggested above

oml/utils/audios.py
AlekseySh: at least we can use DEFAULT_SAMPLE_RATE
but ideally we should pass the actual sample rate
correct me if I'm wrong

oml/utils/download_mock_dataset.py
```diff
@@ -19,54 +23,98 @@
...
     Args:
-        dataset_root: Path to save the dataset
-        check_md5: Set ``True`` to check md5sum
-        df_name: Name of csv file for which output DataFrames will be returned
-        global_paths: Set ``True`` to cancat paths and ``dataset_root``
-
-    Returns: Dataframes for the training and validation stages
-
+        dataset_local_folder: The directory where the dataset will be saved.
+        dataset_remote_folder: The remote directory from which the dataset will be downloaded.
```
AlekseySh, line 40: > The remote directory on `oml.daloroserver.com` from which the dataset will be downloaded.

oml/utils/download_mock_dataset.py
```diff
@@ -19,54 +23,98 @@
...
+    dataset_md5 = dataset_md5 if check_md5 else None
 
-    if not check_exists_and_validate_md5(dataset_root, MOCK_DATASET_MD5 if check_md5 else None):
+    if not check_exists_and_validate_md5(dataset_local_folder, dataset_md5):
         try:
             print("Downloading from oml.daloroserver.com")
-            download_folder_from_remote_storage(remote_folder=MOCK_DATASET_PATH.name, local_folder=str(dataset_root))
-            assert check_exists_and_validate_md5(dataset_root, MOCK_DATASET_MD5 if check_md5 else None)
+            download_folder_from_remote_storage(
+                remote_folder=dataset_remote_folder.name, local_folder=str(dataset_local_folder)
```
AlekseySh, line 63: I think there is a confusing logic in `dataset_remote_folder` argument and the way we use it. Later on, in the domain-specific function we pass **local** path here (for example, `MOCK_DATASET_PATH`, which even starts from local cache folder), after which we only keep the last part of the path calling `.name()` and use it to navigate in server directories.

IMHO, the better option is not to transform `dataset_remove_folder` inside this function and have this logic outside like this:

images:
```
    return _get_mock_dataset(
        dataset_local_folder=dataset_root,
        dataset_remote_folder=MOCK_DATASET_PATH.name,  # <<<< HERE
        dataset_md5=MOCK_DATASET_MD5,
        dataset_gdrive_url=MOCK_DATASET_URL_GDRIVE,
        df_name=df_name,
        check_md5=check_md5,
        global_paths=global_paths,
    )
```
and audio:

```
return _get_mock_dataset(
        dataset_local_folder=dataset_root,
        dataset_remote_folder=MOCK_AUDIO_DATASET_PATH.name,  # <<<<<< HERE
        dataset_md5=MOCK_AUDIO_DATASET_MD5,
        dataset_gdrive_url=MOCK_AUDIO_DATASET_URL_GDRIVE,
        df_name=df_name,
        check_md5=check_md5,
        global_paths=global_paths,
    )
```

AlekseySh (changes requested):
**A few notes about dataframe:**

1. Please, add the description of the start_time column here: https://github.com/OML-Team/open-metric-learning/blob/main/docs/source/oml/data.rst

2. I think it may be confusing that all examples if df.csv within the range of (0, 1). So, before checking the docstring I thought that the measure is a fraction of the audio file. Can we replace a few values with ones greater than 1.0?

3. What if I don't have start_time for some of the records? I suggest to leave some of the records empty in `df_with_start_time.csv`. And add `fillna(0.0)` in your `parse_start_time()` function. We have similar thing for bboxes: some of the images may not have bboxes.

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+        input_tensors_key: str = INPUT_TENSORS_KEY,
+        index_key: str = INDEX_KEY,
+        sr: int = DEFAULT_SAMPLE_RATE,
+        max_num_seconds: float = DEFAULT_DURATION,
+        num_channels: int = DEFAULT_AUDIO_NUM_CHANNELS,
+        start_times: Optional[List[float]] = None,
+        spec_repr_func: Callable[[FloatTensor], FloatTensor] = default_spec_repr_func,
+    ):
+        """
+        Initializes the AudioDataset.
```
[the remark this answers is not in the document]
AlekseySh: The same for other similar comments in init.

oml/utils/audios.py
AlekseySh: also, should we convert the hardcoded values to the function arguments with the default values?

amanteur: > Should this sample_rate=16_000 somehow be synchronized with ours used for model training?

This transform is solely used for visualisation, so it can't be used in training, because user would define own transform, which will be initialised only once somewhere near the dataset/model initialisation and applied during training loop.

i'll add DEFAULT_SAMPLE_RATE

oml/utils/audios.py
amanteur: why cant we use just strings instead of `Tuple[int, int, int]`? Like "read"/"black"/etc? Only for more customisation?

tests/test_oml/test_datasets/test_audio.py
amanteur: yeah, this was dumb from me to use `num_channels`, because there is usually (99%) only two or one channel, and users mostly downmix them into mono audios and leave the same (very rare), so I changed it from `num_channels` to `is_mono`, so it would be more logically right

oml/const.py
```diff
@@ -101,6 +102,17 @@ def get_cache_folder() -> Path:
...
 HYDRA_BEHAVIOUR = "1.1"
+
+# Audio specific constants
+MOCK_AUDIO_DATASET_PATH = CACHE_PATH / "mock_audio_dataset"
+MOCK_AUDIO_DATASET_URL_GDRIVE = "https://drive.google.com/drive/folders/1NcKnyXqDyyYARrDETmhJcTTXegO3W0Ju"
+MOCK_AUDIO_DATASET_MD5 = "b6d1818fe0491909fa5ff8749c82c023"
+
+AUDIO_EXTENSIONS = [".mp3", ".wav", ".ogg", ".flac"]
+START_TIME_COLUMN = "start_time"
+DEFAULT_DURATION = 3.0
```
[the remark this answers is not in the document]
amanteur: fixed

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+def parse_start_times(df: pd.DataFrame) -> Optional[List[float]]:
+    """
+    Parses starting time points from DataFrame.
+    """
+    start_times = None
+    if START_TIME_COLUMN in df.columns:
+        assert (
+            df[START_TIME_COLUMN].dtype == float
+        ), f"Expected dtype of '{START_TIME_COLUMN}' column is `float`, `{df[START_TIME_COLUMN].dtype}` is found."
+        start_times = df[START_TIME_COLUMN].astype(float).tolist()
```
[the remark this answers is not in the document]
amanteur: fixed

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+    """
+
+    def __init__(
+        self,
+        paths: List[Union[str, Path]],
+        dataset_root: Optional[Union[str, Path]] = None,
+        extra_data: Optional[Dict[str, Any]] = None,
+        input_tensors_key: str = INPUT_TENSORS_KEY,
+        index_key: str = INDEX_KEY,
+        sr: int = DEFAULT_SAMPLE_RATE,
```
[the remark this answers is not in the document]
amanteur: fixed

oml/datasets/audios.py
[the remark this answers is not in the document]
amanteur: fixed

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+        Args:
+            paths: List of audio file paths.
+            dataset_root: Base path for audio files.
+            extra_data: Extra data to include in dataset items.
+            input_tensors_key: Key under which audio tensors are stored.
+            index_key: Key for indexing dataset items.
+            sr: Sampling rate of audio files.
+            max_num_seconds: Duration to use from each audio file.
+            num_channels: Number of audio channels.
+            start_times: List of start time offsets in **seconds** for each audio.
```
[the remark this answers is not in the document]
amanteur: fixed

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+            spec_repr_func: Spectral representation extraction function used for visualization.
+        """
+        assert (start_times is None) or (len(paths) == len(start_times))
```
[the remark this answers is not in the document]
amanteur: fixed

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+
+        self.input_tensors_key = input_tensors_key
+        self.index_key = index_key
+
+        self._paths = [Path(p) if dataset_root is None else Path(dataset_root) / p for p in paths]
+        assert all(
+            path.suffix in AUDIO_EXTENSIONS for path in self._paths
+        ), f"Input audios should have one of '{AUDIO_EXTENSIONS}' extensions."
+
+        self.sr = sr
```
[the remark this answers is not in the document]
amanteur: fixed

oml/datasets/audios.py
```diff
@@ -0,0 +1,389 @@
...
+            audio = audio.mean(dim=1, keepdim=True)
+        if sr != self.sr:
+            resampler = Resample(sr, self.sr)
+            audio = resampler(audio)
+        return audio
+
+    def _trim_or_pad(self, audio: FloatTensor, frame_offset: int) -> FloatTensor:
+        """
+        Trim or pad the audio to match the desired number of frames.
+        if `start_time` is specified, then it will be used, else if
```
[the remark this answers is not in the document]
amanteur: fixed

oml/datasets/audios.py
amanteur: fixed

oml/datasets/audios.py
amanteur: fixed

tests/test_oml/test_datasets/test_audio.py
```diff
@@ -0,0 +1,95 @@
...
+    # test non-base dataset
+    df[START_TIME_COLUMN] = [random.uniform(0, 1) for _ in range(len(df))]
+    dataset = AudioLabeledDataset(df)
+    for _ in dataset:
+        pass
+    assert True, "Dataset iteration failed with start times"
+
+
+@pytest.mark.needs_optional_dependency
+def test_audio_datasets() -> None:
```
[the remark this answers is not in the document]
amanteur: fixed

tests/test_oml/test_datasets/test_audio.py
```diff
@@ -0,0 +1,95 @@
+import random
```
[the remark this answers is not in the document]
amanteur: fixed

tests/test_oml/test_datasets/test_audio.py
amanteur: fixed

tests/test_oml/test_datasets/test_audio.py
amanteur: fixed

oml/utils/download_mock_dataset.py
amanteur: fixed

oml/utils/download_mock_dataset.py
amanteur: fixed

oml/utils/audios.py
AlekseySh: yes, don't want to overcomplicate it and support two types of inputs or maintain a mapping from strings to the values

amanteur: I mean that `ax.spines[axis].set_edgecolor()` accepts as input string values like `red`,`black`,etc. Why did you use tuples of ints in the first place?

AlekseySh: when I first introduced colors there was opencv
anyway, I think it's matter of taste 
so, I would rather remove the todo

amanteur: done

oml/utils/audios.py
AlekseySh: hm, okay, I'm not happy with that (I mean user may be confused by seeing a visualised spectogram with the SR different from the one used in his training), but let's leave this as it is in order to finish the PR faster :)

tests/test_oml/test_datasets/test_audio.py
AlekseySh: okay, let's keep this approach

But `is_mono: bool = DEFAULT_IS_MONO` sounds like a flag indicating if an audio is in mono format or not.
I suggest `convert_to_mono: bool = CONVERT_TO_MONO_DEFAULT` which is better semantically from my point of view.

oml/datasets/audios.py
```diff
@@ -0,0 +1,392 @@
...
+        from torchaudio.transforms import Resample
+
+        if self._is_mono and audio.shape[0] != 1:
+            audio = audio.mean(dim=1, keepdim=True)
+        if sample_rate != self._sample_rate:
+            resampler = Resample(sample_rate, self._sample_rate)
+            audio = resampler(audio)
+        return audio
+
+    def _trim_or_pad(self, audio: FloatTensor, frame_offset: int) -> FloatTensor:
```
AlekseySh, line 133: there is a bit of inconsistency here:
- `frame_offset` passed as arguments but it lives in `self._frame_offsets[idx]`
- but `self.num_frames` is read from `self` and not passed as an argument

I suggest to make this function static (or move it out of the class, it's up to you) and pass `num_frames` the same way as you pass `frame_offset`.

oml/datasets/audios.py
```diff
@@ -0,0 +1,392 @@
...
+        Args:
+            df: DataFrame with input data.
+            dataset_root: Base path for audio files.
+            extra_data: Extra data to include in dataset items.
+            input_tensors_key: Key under which audio tensors are stored.
+            index_key: Key for indexing dataset items.
+            labels_key: Key under which labels are stored.
+            sample_rate: Sampling rate of audio files.
+            max_num_seconds: Duration to use from each audio file.
+            is_mono: If True, the audio will be downmixed to one channel; otherwise, it will remain unchanged.
```
AlekseySh, line 249: add symbols `` around True for readthedocs: if `True` ...

oml/datasets/audios.py
```diff
@@ -0,0 +1,392 @@
...
```
AlekseySh, line 363: the same: add ``

oml/utils/audios.py
```diff
@@ -0,0 +1,125 @@
...
+from PIL import Image
+from torch import FloatTensor
+
+from oml.const import BLACK, DEFAULT_SAMPLE_RATE, TColor
+
+
+def default_spec_repr_func(audio: FloatTensor) -> FloatTensor:
+    """
+    Generate a default spectral representation (log-scaled MelSpec) from an audio signal.
+    Used primarily for visualization.
```
AlekseySh, line 17: It's better not to mention how you gonna use the function because it may be changed later or it may push a developer to add a bit more context to the function than it need to know

oml/utils/audios.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _visualize_audio(
+    spec_repr: FloatTensor, color: TColor = BLACK, draw_bbox: bool = True, return_b64: bool = False
+) -> Union[np.ndarray, str]:
+    """
+    Internal function to visualize an audio spectrogram.
+
+    Args:
+        spec_repr: The spectrogram representation as a FloatTensor.
```
AlekseySh, line 40: remove type from here, it's already in typings

oml/utils/audios.py
```diff
@@ -0,0 +1,125 @@
...
+        color: The color of the bounding box.
+        draw_bbox: Whether to draw a bounding box around the spectrogram.
+        return_b64: Whether to return the image as a base64 string.
+
+    Returns:
+        If return_b64 is False, returns the image as a NumPy array.
```
AlekseySh, line 46: please add `` everywhere it's needed in the docs

oml/utils/audios.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def visualize_audio_with_player(
+    audio: FloatTensor, spec_repr: FloatTensor, sample_rate: int, color: TColor = BLACK, draw_bbox: bool = True
+) -> str:
+    """
+    Visualize an audio spectral representation and provide an HTML string with an audio player.
+
+    Args:
+        audio: The audio waveform as a FloatTensor.
```
AlekseySh, line 97: rm types please

oml/utils/audios.py
```diff
@@ -0,0 +1,125 @@
...
+        spec_repr: The spectral representation as a FloatTensor.
```
AlekseySh, line 98: the same

oml/utils/download_mock_dataset.py
```diff
@@ -19,54 +23,97 @@
...
-        global_paths: Set ``True`` to cancat paths and ``dataset_root``
-
-    Returns: Dataframes for the training and validation stages
-
+        dataset_local_folder: The directory where the dataset will be saved.
+        dataset_remote_folder: The remote directory on `oml.daloroserver.com` from which the dataset will be downloaded.
+        dataset_md5: The MD5 checksum used to validate the dataset.
+        dataset_gdrive_url: The Google Drive URL for the dataset download as a fallback option.
+        df_name: The name of the CSV file from which the output DataFrames will be generated.
+        check_md5: If True, validates the dataset using an MD5 checksum. Default is True.
```
AlekseySh, line 44: add `` around True

oml/utils/download_mock_dataset.py
```diff
@@ -19,54 +23,97 @@
...
+        global_paths: If True, concatenates the paths in the dataset with the dataset_local_folder. Default is False.
```
AlekseySh, line 45: the same

oml/utils/download_mock_dataset.py
```diff
@@ -19,54 +23,97 @@
...
+    check_md5: bool = True,
+    global_paths: bool = False,
+) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    """
+    Function to download mock images dataset which is already prepared in the required format.
+
+    Args:
+        dataset_root: The directory where the dataset will be saved.
+        df_name: The name of the CSV file from which the output DataFrames will be generated.
+        check_md5: If True, validates the dataset using an MD5 checksum. Default is True.
```
AlekseySh, line 98: add `` around True

oml/utils/download_mock_dataset.py
```diff
@@ -19,54 +23,97 @@
...
+        global_paths: If True, concatenates the paths in the dataset with the dataset_local_folder. Default is False.
```
AlekseySh, line 99: add `` around True

tests/test_oml/test_datasets/test_datasets.py
```diff
@@ -141,3 +151,74 @@ def test_image_datasets() -> None:
     check_base(dataset_qgl)
     check_query_gallery(dataset_qgl, df_val)
     check_labeled(dataset_qgl, df_val)
+
+
+def get_df() -> pd.DataFrame:
```
AlekseySh, line 156: hm, you probably understood me wrong on moving tests into another py file, I'm sorry if it was not clear enough. The idea was:

`def test_audio_datasets(df: pd.DataFrame) -> None:` - keep it in `test_datasets` (as it's done now) because we don't test domain specific stuff, we just check that all audio datasets are aligned with the general interfaces

add `test_audio_datasets.py` file for domain specific stuff and put `test_downmix`, `test_resmaple_trim_pad`, `test_start_times` there in order not to mix it up with testing general functionality.

oml/datasets/audios.py
```diff
@@ -0,0 +1,392 @@
...
+
+        Args:
+            audio: Audio tensor.
+            frame_offset: Frame offset for trimming.
+
+        Returns:
+            Trimmed and/or padded audio tensor.
+        """
+        if audio.shape[1] > self._num_frames:
+            audio = audio[:, frame_offset : frame_offset + self._num_frames]
```
AlekseySh, line 145: If you s`tart_time` is bigger than a length of an audio, you get array of zeros here, right?
The problem if this condition is met for 2 or more audios. In such case you have a set of the same inputs (which are simply zeros) having some different labels, which will negatively affect the training process.

Thus, I suggest to raise a error instead and add a test on that

AlekseySh (changes requested):
Looks good, almost done :)

oml/datasets/audios.py
amanteur: done

tests/test_oml/test_datasets/test_datasets.py
amanteur: done

oml/utils/download_mock_dataset.py
amanteur: done

oml/utils/download_mock_dataset.py
amanteur: done

oml/utils/download_mock_dataset.py
amanteur: done

oml/utils/audios.py
amanteur: done

oml/utils/audios.py
amanteur: done

oml/datasets/audios.py
amanteur: done

oml/datasets/audios.py
amanteur: done

oml/datasets/audios.py
amanteur: done

oml/utils/audios.py
amanteur: Done

oml/utils/audios.py
amanteur: Done

oml/utils/audios.py
amanteur: Done

oml/utils/download_mock_dataset.py
amanteur: Done

oml/const.py
```diff
@@ -101,6 +102,17 @@ def get_cache_folder() -> Path:
...
 
-# hydra provides ability to set its behaviour for convinient backwards compatibility
+# hydra provides ability to set its behaviour for convenient backwards compatibility
 # no matter what the current version is
 HYDRA_BEHAVIOUR = "1.1"
+
+# Audio specific constants
+MOCK_AUDIO_DATASET_PATH = CACHE_PATH / "mock_audio_dataset"
+MOCK_AUDIO_DATASET_URL_GDRIVE = "https://drive.google.com/drive/folders/1NcKnyXqDyyYARrDETmhJcTTXegO3W0Ju"
+MOCK_AUDIO_DATASET_MD5 = "d983fb30e6cbcf0f79aaf88d2ea0ec83"
```
AlekseySh, line 112: I got a different hash: `45629afad9ff28f9148f3ea60f659d18`

AlekseySh (changes requested):

oml/const.py
DaloroAT: For me `45629afad9ff28f9148f3ea60f659d18` too with updated structure

AlekseySh: @DaloroAT Aman said it was `.DS_Store` file locally)

DaloroAT: Haha 😁 💩

PR #608, opened 2024-06-25
Update README.md

I8dNLo:
get_treansforms_for_pretrained can not be imported from oml.registry, but can be imported from oml.registry.transforms.

For maxi

- [x] I've checked contribution [guide](https://open-metric-learning.readthedocs.io/en/latest/oml/contributing.html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,7 +593,7 @@ from oml.losses import TripletLossWithMiner
 from oml.metrics import calc_retrieval_metrics_rr
 from oml.miners import AllTripletsMiner
 from oml.models import ViTExtractor
-from oml.registry import get_transforms_for_pretrained
+from oml.registry.transforms import get_transforms_for_pretrained
 from oml.retrieval import RetrievalResults, AdaptiveThresholding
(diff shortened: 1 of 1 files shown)


AlekseySh:
Hi, @I8dNLo and thank you for using OML!

There is a shortcut for the import in `__init__`: https://github.com/OML-Team/open-metric-learning/blob/main/oml/registry/__init__.py#L9.
So, it should work.

We also run code examples in CI system.

What type of error did you get?

I8dNLo:
Yap, it should work. But it does not :0 
I just manually tried your installation guidelines. Also [nlp] version does not install too. 
But if someone would import it like I did, all the thing will work

AlekseySh:
@I8dNLo it's super weird. If you open an example in colab it will also work...

PS. The reason why I don't want to instantly accept the changes is because I try to decrease the width of the two-column code examples in Readme so we could see both (at least partially).

AlekseySh:
@DaloroAT what do u think? why it may happen?

AlekseySh:
due to inactivity

PR #609, opened 2024-06-26
Refactored datasets classes

AlekseySh:
There are a few issues with the current implementations of the dataset interfaces:

* There is a lot of duplicated code related to dataframe parsing and processing. We have the same code two times (texts, images) and are going to repeat the same thing 3d time for audio. It's too much, so, the classes have to be refactored.
* There is a tricky implementation of `IQueryGalleryDataset` which also needs to be updated.

This PR has to address all of these issues. 

With the proposed changes we also almost avoid inheritance from non-abstract classes. Formally, we still inherit `ImageLabeledDataset` from `LabeledDatasetDF`, but it's rather an alternative constructor than meaningful inheritance with some logic on top of that. (I could literally make `@classmethod`: `create_for_images()` and `create_for_texts()` in `LabeledDatasetDF`, but I need to keep back compatibility).

### DIFF
--- a/oml/datasets/dataframe.py
+++ b/oml/datasets/dataframe.py
@@ -20,8 +20,8 @@ from oml.interfaces.datasets import (
 )
 
 
-def update_dataset_extra_data(dataset, df, extra_data):
-    extra_data = dict() or extra_data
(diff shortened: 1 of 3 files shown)


amanteur:
lgtm!

DaloroAT (approved):
LSFM!
image

PR #615, opened 2024-07-04
Reproduction example

I8dNLo:
Conversion of vitb16_unicom does not work after last release :( 

Oml version: 3.1.3
Torch version: 2.1.1

`RuntimeError: unordered_map::at: key not found
`

### DIFF
--- a/conversion/unicom_to_onnx.py
+++ b/conversion/unicom_to_onnx.py
@@ -1,0 +1,34 @@
+import onnx
+import onnx.checker
+import torch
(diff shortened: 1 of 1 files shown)


AlekseySh:
@I8dNLo oops, we've added the key to the constructor (`__init__`), but forgot about `from_pretraned`
I've released a fix in OML 3.1.4, it will be available in PYPI soon 

by the way, issue is a better format for reporting such stuff than PR

PR #623, opened 2024-07-28
[WIP] Add datasets to config api V2

leoromanovich:
- [ ] I've checked contribution [guide](https://open-metric-learning.readthedocs.io/en/latest/oml/contributing.html).

### DIFF
--- a/oml/configs/datasets/image_labeled_dataset.yaml
+++ b/oml/configs/datasets/image_labeled_dataset.yaml
@@ -1,0 +1,9 @@
+name: image_labeled_dataset
+args:
+  dataframe_name: df.csv
+  cache_size: 0
+  show_dataset_warnings: True
+  transform:
+    name: augs_hypvit_torch
+    args:
+      im_size: 224
--- a/oml/registry/datasets.py
+++ b/oml/registry/datasets.py
@@ -1,0 +1,43 @@
+import inspect
+from pathlib import Path
+from typing import Optional
+
+import pandas as pd
+
+from oml.const import LABELS_COLUMN, SPLIT_COLUMN, TCfg
+from oml.datasets.images import (
+    ImageLabeledDataset,
+    ImageQueryGalleryDataset,
+    ImageQueryGalleryLabeledDataset,
+)
+from oml.interfaces.datasets import IBaseDataset
+from oml.utils.dataframe_format import check_retrieval_dataframe_format
+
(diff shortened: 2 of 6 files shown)


oml/configs/datasets/image_labeled_dataset.yaml
```diff
@@ -0,0 +1,9 @@
+name: image_labeled_dataset
+args:
+  dataframe_name: df.csv
```
AlekseySh, line 3: let's use "df" name here? so it's more consistent with argument names, but replace it with read actual df in runtime

oml/registry/datasets.py
```diff
@@ -0,0 +1,43 @@
...
+from oml.utils.dataframe_format import check_retrieval_dataframe_format
+
+DATASETS_REGISTRY = {
+    "image_labeled_dataset": ImageLabeledDataset,
+    "image_query_gallery_labeled_dataset": ImageQueryGalleryLabeledDataset,
+    "image_query_gallery_dataset": ImageQueryGalleryDataset,
+}
+
+
+def get_dataset_by_cfg(cfg: TCfg, split: Optional[str] = None) -> IBaseDataset:
```
AlekseySh, line 23: I suggest to rework this function a bit:


```
def get_dataset_by_cfg(cfg: TCfg, split: Optional[str] = None) -> IBaseDataset:
    if cfg['name'] in DATASETS_REGISTRY:
        df = pd.read_csv(Path(cfg["args"]["dataset_root"]) / cfg["args"]["df"], index_col=False)

        mapper = {l: i for i, l in enumerate(df.sort_values(by=[SPLIT_COLUMN])[LABELS_COLUMN].unique())}
        df[LABELS_COLUMN] = df[LABELS_COLUMN].map(mapper)

        if split is not None:
            df = df[df[SPLIT_COLUMN] == split].reset_index(drop=True)

        cfg["args"]["df"] = df

    else:
        if split is not None:
            raise ValueError("We only support <split> option for built-in datasets.")

    if split and "dataframe_name" in cfg["args"].keys():

    return dataset_class(**cfg["args"])
```


I also removed `check_retrieval_dataframe_format` because there was no such functionality before 
I also hope in my implementation we dont need filtering arguments

AlekseySh (changes requested):
I like this approach. Let's continue with this. 

As the next step, I think you need to update pipelines code with usage of a new `split` argument

